PR #3, opened 2024-01-31
feat: shared repo config

kylemcd:
- Adds eslint config from `@knocklabs/eslint-config` + corresponding GitHub actions workflow
- Adds typescript config from `@knocklabs/typescript-config`
- Adds prettier config from `@telegraph/prettier-config` + corresponding GitHub actions workflow
  - Might be worth thinking about adding the prettier config to `@knocklabs/prettier-config` but also didn't want to create formatting issues with inflight work in that repo.
- Modifies `.husky` to include `yarn format`


.github/workflows/format.yml
```diff
@@ -0,0 +1,26 @@
+name: Format Project
```
connorlindsey, line 1: Do we want these as separate actions or a single ci/cd one? We've generally had a single one, but I think splitting them is a good option. I imagine this'll make the output clearer in each PR

connorlindsey (approved):

.github/workflows/format.yml
kylemcd: Yeah I appreciate the more granular output. I could move it into one job like this to get more re-use out of `build` / `install` + get the granular output. Did this [here](https://github.com/interunit/ui/blob/release/v0.1.0/.github/workflows/pr-checks.yml) previously. 

Will follow up with a PR addressing this feedback once I have a clearer idea of what I want to be running in Actions.

PR #31, opened 2024-02-06
feat: tailwind config generated from design tokens

kylemcd:
`@telegraph/tokens`
- Adds `generate-css-var-map` script in `@telegraph/tokens` for use in tailwind-config
- Breaks apart tokens in `font` for more granular control over mappings
- Updates build process to use vite for more control over file splitting

`@telegraph/tailwind-config`
- Creates `default` tailwind config which includes mapping to both tokens and semantic tokens
- Created `tokens` tailwind config which includes mapping to just the tokens config
- Updates build process to use vite for more control over file splitting

`@telegraph/postcss-config`
- Updates reference to `@telegraph/tailwind-config` as the export for the config has changed


packages/tokens/scripts/generate-css-var-map.js
```diff
@@ -0,0 +1,119 @@
...
+        !Array.isArray(obj[key])
+      ) {
+        // Recurse into nested objects
+        traverseObject(obj[key], currentPath);
+      } else {
+        // Process leaf nodes
+        if (!currentPath.includes("dark")) {
+          const filteredPath = currentPath
+            .filter((t) => t !== "dark" && t !== "light" && t)
+            .join("-");
```
cjbell, line 37: Nit: I'd probably have this as `t && !["light", "dark"].includes(t)` and even consider extracting that array into a const

packages/tokens/scripts/generate-css-var-map.js
```diff
@@ -0,0 +1,119 @@
...
+const { writeFile, mkdir } = require("node:fs/promises"); // For file system operations (async)
+
+// Constant for the prefix used in CSS variables
+const TELEGRAPH_VARIABLE_PREFIX = "tgph";
+
+/**
+ * Maps CSS variables to class names based on the provided object structure.
+ * @param {Object} obj - The object containing design tokens.
+ * @param {String} prefix - Prefix for the class name.
+ * @param {Boolean} noVariableNamePrefix - Flag to include/exclude the prefix in variable names.
```
cjbell, line 13: Tiny nit: I think it might be better to flip this to have it as "includeVariableNamePrefix" where this defaults to true. Then you don't end up doing the double not `!no` in the one place we use it.

cjbell (approved):
Super cool 💯 – 2 nits, not blocking

PR #32, opened 2024-02-06
feat: heading component

kylemcd:
`@telegraph/typography`
- Adds `<Heading/>` component
- Adds `props-mapping` helper to be used across both `<Heading/>` and `<Text/>`
- Adds `clsx` package for managing classNames

`@telegraph/tokens`
- Adds font weight tokens 

`@telegraph/tailwind-config`
- Adds font weight tokens to tailwind preset

`@telegraph/playground`
- Adds heading variants for viewing


packages/typography/src/Heading/Heading.tsx
```diff
@@ -1,9 +1,39 @@
...
+    {
+      align = "DEFAULT",
+      color = "DEFAULT",
+      size = "DEFAULT",
+      ...props
+    }: HeadingProps,
+    forwardedRef,
+  ) => {
+    return (
+      <h3
```
cjbell, line 25: Feels like an oversight to only have this as an `h3` which then makes me feel like we need an `as` prop which opens up a big-old-can-of-worms.

examples/playground/app/page.tsx
```diff
@@ -3,8 +3,26 @@ import { Heading, Text } from "@telegraph/typography";
...
+      <Heading size="9">Heading</Heading>
+      <Heading size="8">Heading</Heading>
+      <Heading size="7">Heading</Heading>
+      <Heading size="6">Heading</Heading>
+      <Heading size="5">Heading</Heading>
+      <Heading size="4">Heading</Heading>
+      <Heading size="3">Heading</Heading>
+      <Heading size="2">Heading</Heading>
+      <Heading size="1">Heading</Heading>
+      <Heading size="0">Heading</Heading>
```
cjbell, line 15: Few qs:

1. Should size be a string over an integer?
2. Is this what we've agreed upon with Krisna to describe these? I assume so, but asking as last time I thought we had some token names

packages/typography/src/Heading/Heading.tsx
```diff
@@ -1,9 +1,39 @@
...
+  size?: keyof typeof sizeMap;
+  color?: keyof typeof colorMap;
+};
+
+type HeadingRef = HTMLHeadingElement;
+
+const Heading = React.forwardRef<HeadingRef, HeadingProps>(
+  (
+    {
+      align = "DEFAULT",
```
cjbell, line 17: Nit: can we put this const as a const we share between here and the prop mapping

cjbell (commented):
few notes

packages/typography/src/Heading/Heading.tsx
kylemcd: I thought about making the `as` prop a required field here so that the engineer has to decide which heading level to use. But as it exists now there can be an as prop passed optionally. What do you think about making it required?

examples/playground/app/page.tsx
kylemcd: 1. I could make it either but typing `size="2"` is a lot more satisfying than typing `size={2}`
3. Yes this matches the figma doc here: https://www.figma.com/file/br4Lk9C5qw3UVyGsAbONDY/Telegraph-design-system?node-id=38%3A327&mode=dev

packages/typography/src/Heading/Heading.tsx
kylemcd: Not sure I'm following?

You want `const DEFAULT = "DEFAULT"` declared?

cjbell: yep exactly and then use in the prop mappings as `[DEFAULT]: "foo"`

packages/typography/src/Heading/Heading.tsx
cjbell: I think making it required is good because it forces you to think about the heading level. Could see everything being an h3 otherwise. I'm an old school semantic HTML guy though so maybe need to take this through that lens

cjbell (approved):
💯

packages/typography/src/Heading/Heading.tsx
kylemcd: I can make it required, I was going to initially but that might be too prescriptive. Will update!

packages/typography/src/Heading/Heading.tsx
cjbell: Oh yeah that's even better!

PR #34, opened 2024-02-07
feat: text component

kylemcd:
`@telegraph/typography`
- Fixes type issues in existing tests
- Fixes letting the `className` prop through to the component
- Adds the Text Component

### DIFF
--- a/packages/typography/src/Text/Text.tsx
+++ b/packages/typography/src/Text/Text.tsx
@@ -1,5 +1,59 @@
-function Text() {
-  return <p className="text-amber-400">Text</p>;
-}
+import clsx from "clsx";
+import React from "react";
+
+import {
+  alignMap,
+  colorMap,
+  sizeMap,
+  weightMap,
+} from "../helpers/prop-mappings";
+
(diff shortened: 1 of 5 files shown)


packages/typography/src/Text/Text.tsx
```diff
@@ -1,5 +1,59 @@
...
+  color?: keyof typeof colorMap;
+  weight?: keyof typeof weightMap;
+};
+
+type TextRef = HTMLElement;
+
+const Text = React.forwardRef<TextRef, TextProps>(
+  (
+    {
+      color = "black",
```
connorlindsey, line 34: Do we want to allow for using other colors from the design system, e.g. a lighter shade of gray or do we always want to restrict it to the ones defined in colormap?

connorlindsey (approved):
Looks great! Just one question re. what colors we want to allow

packages/typography/src/Text/Text.tsx
kylemcd: We want to restrict it to those defined in the color map

PR #40, opened 2024-02-09
feat: next.js plugin for telegraph

kylemcd:
When utilizing useContext in the new button component, Next.js (RSC) would complain about not having a `"use client"` directive in the file. Because `"use client"` is only currently in Next.js including this directive generally in the repo wouldn't make sense. So this work creates a Next.js plugin that applies the `"use client"` directive onto telegraph packages.

### DIFF
--- a/examples/playground/use-client-loader.js
+++ b/examples/playground/use-client-loader.js
@@ -1,0 +1,3 @@
+module.exports = function (source) {
+  return `"use client";\n` + source;
+};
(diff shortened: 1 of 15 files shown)


examples/playground/use-client-loader.js
```diff
@@ -0,0 +1,3 @@
+module.exports = function (source) {
+  return `"use client";\n` + source;
+};
```
cjbell, line 3: I don't think you need this in the playground now?

cjbell (approved):
one comment but looks good!

connorlindsey (approved):

PR #43, opened 2024-02-12
feat: updated leading tokens & surface color tokens

kylemcd:
- The leading tokens were copy pasta-ed incorrectly so those are now corrected to match figma
- There was additional colors added for surface colors


packages/tokens/src/tokens/color.ts
```diff
@@ -3,8 +3,14 @@ export const color = {
...
-    light: "#ffffff",
-    dark: "#111113",
+      1 :{
+          light: "#ffffff",
+          dark: "#18191b",
+      },
+      2 :{
+    light: "#f9f9f8",
+    dark: "#111110",
+      }
```
cjbell, line 13: Formatting looks whacky here, is prettier setup for the build in CI?

cjbell (approved):
one nit, not blocking

packages/tokens/src/tokens/color.ts
kylemcd: Yeah in CI and on pre-commit. No idea how that go through 🤔

PR #47, opened 2024-02-12
feat: button component

kylemcd:
`@telegraph/button`
- Initial button buildout

Note: something went wrong with a rebase so a bunch of commits are here that are outside of the scope of the PR. The files changed are all correct though.


packages/button/src/Button/Button.tsx
```diff
@@ -0,0 +1,184 @@
...
+    React.useEffect(() => {
+      const children = React.Children.toArray(props.children);
+      if (children?.length === 1 && React.isValidElement(children[0])) {
+        const child = children[0] as React.ReactComponentElement<typeof Icon>;
+        if (child?.type?.displayName === "Icon") {
+          return setLayout("icon-only");
+        }
+        return setLayout("default");
+      }
+    }, [props.children]);
```
kylemcd, line 75: Open to feedback on a better way to do this. Could do it with css but don't want to muddy up how the tokens are setup for size. This also throws a warning in RSC as the className doesn't match first render. Maybe a ref could be good?

packages/button/src/Button/Button.constants.ts
```diff
@@ -0,0 +1,111 @@
...
+    accent: "accent",
+    gray: "gray",
+    red: "red",
+    disabled: "disabled",
+  },
+} as const;
+
+export const iconVariantMap = {
+  default: "secondary",
+  "icon-only": "primary",
```
thomaswhyyou, line 110: Sorry if I should know this already - what is the difference between "primary" vs "secondary" icon variants?

thomaswhyyou (commented):
Took an initial pass and left some questions/comments for `Button.tsx` and `Button.constants.tsx`. The main thing that kinda stuck out to me was around having handle "layout" vs "icon-only" in a couple different places.

packages/button/src/Button/Button.tsx
```diff
@@ -0,0 +1,202 @@
...
+import {
+  buttonColorMap,
+  buttonSizeMap,
+  iconColorMap,
+  iconSizeMap,
+  iconVariantMap,
+  textColorMap,
+  textSizeMap,
+} from "./Button.constants";
+
```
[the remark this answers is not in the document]
kylemcd: Here you go: https://www.figma.com/file/br4Lk9C5qw3UVyGsAbONDY/Telegraph-design-system?type=design&node-id=63-9651&mode=design&t=3BfAH0suMUoYAuPz-0

packages/button/src/Button/Button.tsx
```diff
@@ -0,0 +1,202 @@
...
+type RootBaseProps = {
+  variant?: "solid" | "soft" | "outline" | "ghost";
+  size?: "1" | "2" | "3";
+  color?: "accent" | "gray" | "red" | "disabled";
```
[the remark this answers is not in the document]
kylemcd: Yeah disabled is a state but there is a color scheme tied to it (see figma), so it needs to exist alongside the other colors config. You're able to pass the `disabled` prop like normal into the button and it'll work.

packages/button/src/Button/Button.constants.ts
```diff
@@ -0,0 +1,111 @@
...
+export const buttonSizeMap = {
+  default: {
+    "1": "h-6 px-2 gap-1",
+    "2": "h-8 px-3 gap-2",
+    "3": "h-10 px-4 gap-3",
+  },
+  "icon-only": {
+    "1": "h-6 w-6",
+    "2": "h-8 w-8",
+    "3": "h-10 w-10",
```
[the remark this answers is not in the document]
kylemcd: With how spacing was setup in the button in figma letting the padding control the width wouldn't make it a perfect square. So this fixes that issue without introducing a ton of one off spacing handling for icon only styles that would need to be done elsewhere.

packages/button/src/Button/Button.constants.ts
kylemcd: Primary is a darker color and secondary is a slightly lighter color. This is documented in the figma file as well.

packages/button/src/Button/Button.tsx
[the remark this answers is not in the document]
kylemcd: Accent is the brand color. We're following the same naming convention as Radix Themes. Brand felt too prescriptive of a name where as accent makes more sense. So for us, this would be the knock orange. You can see all of the config for these colors here: https://github.com/knocklabs/telegraph/blob/main/packages/tokens/src/tokens/color.ts

packages/button/src/Button/Button.tsx
thomaswhyyou: Yeah this doesn't feel super clean, but based on the current design system it seems like the choice is to either handle the "icon only" styles with one off CSS, or with this. i.e. no easy way to avoid that particular complexity, and at the end of the day it's not very consequential as long as it renders the right thing, given it's contained in this component. 

I guess my only other thought/question — do we need to do this with `useState`+`useEffect` though? This is the only place where this state is set, which makes me think we can just use `useMemo`?

thomaswhyyou (commented):
A couple small follow up thoughts/questions.

packages/button/src/Button/Button.constants.ts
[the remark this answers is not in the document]
kylemcd: Yeah I agree, it's definitely a smell but I don't think it has any adverse side effects so probably fine for now. If we run into issues later maybe we change it.

packages/button/src/Button/Button.constants.ts
kylemcd: Yep just a design decision

packages/button/src/Button/Button.tsx
kylemcd: Yeah I'm going to take another stab at removing the useEffect in exchange for something else

packages/icon/src/Icon/Icon.tsx
```diff
@@ -27,16 +27,29 @@ const Icon = React.forwardRef<IconRef, IconProps>(
...
+
+    // Replace ionicon classes with telegraph classes
+    const iconMarkup = unsafeIconMarkup
+      ?.replace(/class='ionicon'/g, "")
+      ?.replace(/ionicon-stroke-width/g, "tgph-stroke-width")
+      ?.replace(/ionicon-fill-none/g, "tgph-fill-none");
+
+    if (!iconMarkup) {
+      return <></>;
+    }
```
thomaswhyyou, line 41: Curious of the reason that is driving the changes in this [commit](https://github.com/knocklabs/telegraph/pull/47/commits/2b93350b465ad67bbcad9a2c9614ba4a8803d7c2)? Not sure if I fully follow - we are using Ion icons, but not with `IonIcon`? Maybe that's something we want to explain/document more in the comment above?

thomaswhyyou (approved):
One question regarding the changes around the Icon component. Not sure what's happening there, and I don't have enough contexts to provide feedbacks. 

Doesn't feel like a blocker though, and everything else looks good to me!

packages/icon/src/Icon/Icon.tsx
kylemcd: The icon that I was originally using from `@ionic/react` isn't compatible with RSC because it adds to the className prop internally outside of React's lifecycle, this was causing hydration issues. I realized that once digging into the `IonIcon` component that it's doing way more than we need it to. So, instead of dragging around a heavy-ish external library to display the icon we already get from the `ionicons` package being passed in as the `icon` prop, we can just handle displaying it ourselves. So this component now grabs the svg markup from the `icon` prop passed in, adds our classes to it to avoid the scenario of collision, and display it that way. Doing this also means that it's within the React Lifecycle properly and is compatible with RSC.

thomaswhyyou: Cool, thanks for explaining!

PR #48, opened 2024-02-14
feat: scoped telegraph classes

kylemcd:
In order to avoid collisions between telegraph tailwind classes and existing tailwind classes in a project we've created `scoped` versions of stylesheets that can be optionally used.

> See changeset files for specific package changes


packages/tailwind-config/src/scoped-config.ts
```diff
@@ -0,0 +1,6 @@
+import defaultConfig from "./default-config";
+
+export default {
+  important: ".tgph",
```
thomaswhyyou, line 4: I don't know enough about vite/monorepo configs to provide any detailed feedbacks here (maybe @connorlindsey can skim over this?), but one high level question I had was — is there a reason we don't want to just scope by default? 

I'm guessing it makes no difference for our internal use, but for third party usage it seems reasonable and simpler to just scope it by .tgph namespace by default? Not sure if there's any adverse side effects i'm not thinking of.

connorlindsey: I think that's a good question. Should we scope all styles by default? I haven't looked too deeply into this PR yet but if it's not too much of a DX pain in the design system, might be worth considering

thomaswhyyou (approved):
To be honest I don't grok all the configs here, but approving with the assumption/trust that this does what it needs to based on our discussions in Slack! 🙂

PR #49, opened 2024-02-15
fix: remove external packages from bundles

kylemcd:
We were including external dependencies in our bundles which was making our bundle size huge. This PR modifies our vite config to exclude third party dependencies from our bundles.

### DIFF
--- a/packages/vite-config/src/vite-config.ts
+++ b/packages/vite-config/src/vite-config.ts
@@ -1,19 +1,46 @@
 import react from "@vitejs/plugin-react";
+import { createRequire } from "node:module";
+import { resolve } from "path";
 import dts from "vite-plugin-dts";
+
+const require = createRequire(import.meta.url);
+const pkg = require(resolve(process.cwd(), "package.json"));
+
+// Get package name so we can use it as the library name
+const packageName = pkg.name;
+
+// Get all dependencies so we can tell vite not to bundle them with our packages
+const dependencies = Object.keys(pkg.dependencies || {});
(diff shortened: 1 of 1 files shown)


packages/vite-config/src/vite-config.ts
```diff
@@ -1,19 +1,46 @@
...
+const dependencies = Object.keys(pkg.dependencies || {});
+const peerDependencies = Object.keys(pkg.peerDependencies || {});
+
+const allDependencies = [
+  ...dependencies,
+  ...peerDependencies,
+  // Need to declare these as external as well since they're
+  // not explicitly listed in the package.json
+  "react/jsx-runtime",
+  "ionicons/icons",
```
thomaswhyyou, line 22: This looks a little curious because I thought we are using ionic icons for the Icon component, so I would've thought it should be one of the dependencies? Would you mind explaining a bit more here for my info? 

Just to trying to understand why some of these dependencies need to be manually set. And wondering if/how often this is something that we will constantly have to remember to list manually as we evolve the the design system package.

kylemcd: It is a dependency, it's just removing `ionicons` from the bundle since `ionicons` will be installed along with `@telegraph/icons` as it's a dependency. 

The manual ones are set because `ionicons/icons` and `react/jsx-runtime` are not explicitly listed in the package json as it has a `/` in it. Ideally I'd find a better way to handle these but think it's good enough for now.

thomaswhyyou (approved):

PR #52, opened 2024-03-01
fix: esm build issues + build cleanup

kylemcd:
- Fixes the same issue we saw in `knocklabs/javascript` where `.es.js` files would not resolve in a module as the correct file extension should be `.mjs`
- Simplifies vite build config
- Separates out dist files into folder like `knocklabs/javascript`
- Simplifies build for packages into one `vite build`
- Updates `package.json` exports to new paths

### DIFF
--- a/.changeset/selfish-shirts-rescue.md
+++ b/.changeset/selfish-shirts-rescue.md
@@ -1,0 +1,11 @@
+---
+"@telegraph/tailwind-config": patch
+"@telegraph/vite-config": patch
+"@telegraph/typography": patch
+"@telegraph/button": patch
+"@telegraph/nextjs": patch
+"@telegraph/tokens": patch
+"@telegraph/icon": patch
+---
+
+fixes esm build issues, removes semantic tokens, deprecates the core package


packages/tailwind-config/src/default-config.ts
```diff
@@ -1,23 +1,18 @@
-import semantic from "@telegraph/tokens/mappings/semantic";
-import tokens from "@telegraph/tokens/mappings/tokens";
+import tokens from "@telegraph/tokens/css-variables-map";
```
thomaswhyyou, line 1: Question for my understanding: so `"@telegraph/tokens/css-variables-map"` now has combined mappings from both `"@telegraph/tokens/mappings/semantic"` and `"@telegraph/tokens/mappings/tokens"`?

Asking because I wasn't sure where that combining happens in this PR, or maybe that was already the case before.

thomaswhyyou (approved):
I confess i don't understand all the config changes here, but they look legit? Also I realize it's a low risk in tweaking configs for our needs since we are the only user for now.

packages/tailwind-config/src/default-config.ts
kylemcd: Ah, should have called out that I'm removing the semantic tokens in a follow up PR. Just dropped it here to create less work for this PRs focus.

thomaswhyyou: Got it! 👍

PR #54, opened 2024-03-01
chore: remove semantic tokens

kylemcd:
This was a concept we thought we'd use but it ended up manifesting itself inside of the components themselves. So this deletes these files as they're not needed and cleans up references to "semantic".


thomaswhyyou (approved):
Approving, per our [prior discussion](https://github.com/knocklabs/telegraph/pull/52#discussion_r1509515577)!

PR #60, opened 2024-03-05
feat: form field and text field

kylemcd:
`@telegraph/text-field`
- Adds component

`@telegraph/form`
- Adds Field component

`@telegraph/composed-refs`
- adds composed refs util to merge forwarded refs and internal refs. Take from Radix: https://github.com/radix-ui/primitives/tree/main/packages/react/compose-refs

`@telegraph/button`
- Adds data attributes for better targeting, going to add these to all of our components in a follow up PR
- Fixed outlined button border

`@telegraph/icon`
- Adds data attributes


examples/playground/app/page.tsx
```diff
@@ -1,11 +1,54 @@
...
+      <div style={{ maxWidth: "500px", width: "100%" }}>
+        <Field
+          size="1"
+          label="Label"
+          placeholder="Placeholder"
+          value={value}
+          onChange={setValue}
+          leadingIcon={{ icon: searchOutline, alt: "add icon" }}
+          TrailingAction={<Button icon={{ icon: addSharp }} />}
+        />
```
kylemcd, line 22: What a `<Field/>` component looks like in action.

packages/button/src/Button/Button.tsx
```diff
@@ -84,7 +84,11 @@ const Root = React.forwardRef<RootRef, RootProps>(
             buttonSizeMap[layout][size],
             className,
           )}
-          data-button-layout={layout}
+          data-tgph-button
+          data-tgph-button-layout={layout}
+          data-tgph-button-size={size}
+          data-tgph-button-variant={variant}
+          data-tgph-button-state={color}
```
kylemcd, line 91: data attributes for css targeting

packages/compose-refs/src/compose-refs/compose-refs.ts
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * A custom hook that composes multiple refs
+ * Accepts callback refs and RefObject(s)
+ */
+function useComposedRefs<T>(...refs: PossibleRef<T>[]) {
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  return React.useCallback(composeRefs(...refs), refs);
+}
+
+export { composeRefs, useComposedRefs };
```
kylemcd, line 35: Utility used to compose refs. Since we forwardRefs to our components, if we need to also have a ref internally (like in `<TextField/>`), these need to be merged so you can apply one ref to the element.

packages/form/src/Field/Field.tsx
```diff
@@ -0,0 +1,197 @@
...
+        <Control>
+          {control.name === "TextField" && (
+            <TextField {...props} {...control.props} />
+          )}
+        </Control>
+      )}
+      {(message || typeof error === "string") && <Message>{message}</Message>}
+    </Root>
+  );
+};
```
kylemcd, line 181: The `<Field/>` component. Let's you pass in normal things like `label`, `value`, `onChange`. Also lets you decide what form control to use (just text field for now). The `<Control/>` component wraps this so that we can pass values directly to the chosen form control via a [slot](https://www.radix-ui.com/primitives/docs/utilities/slot) in a modular way.

packages/text-field/src/TextField/TextField.tsx
```diff
@@ -0,0 +1,135 @@
...
+            return;
+          }
+
+          const input = inputRef.current;
+          if (!input) return;
+
+          requestAnimationFrame(() => {
+            input.focus();
+          });
+        }}
```
kylemcd, line 100: This handles 2 things. 
- When there is a trailing action button that is clicked, the field doesn't pick up that focus state so the field remains in it's normal state
- When there is an icon in the front without an action, when the icon is clicked the field is focused for the user.

packages/text-field/src/TextField/TextField.tsx
```diff
@@ -0,0 +1,135 @@
...
+          <div
+            className={clsx(
+              "box-border aspect-square h-full p-[2px] flex items-center justify-center",
+              // Overrides to get the button to fit into the text field as designed
+              "[&>[data-tgph-button]]:aspect-square [&>[data-tgph-button]]:p-0 [&>[data-tgph-button]]:w-full [&>[data-tgph-button]]:h-auto",
+              SIZE.TrailingAction[size],
+            )}
+          >
+            {TrailingActionComponent}
+          </div>
```
kylemcd, line 128: This sets some overrides to the Telegraph Button in order to get it to match the desired design. Notice how we're ONLY targeting the direct child `[data-tgph-button]` in order to not override other elements.

packages/text-field/src/TextField/TextField.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+    // Adds default props to the component, which can be overridden by the caller
+    const TrailingActionComponent = TrailingAction
+      ? React.cloneElement(TrailingAction, {
+          size,
+          color: "gray",
+          variant: "ghost",
+          ...TrailingAction?.props,
+        })
+      : null;
```
kylemcd, line 75: This takes in the `TrailingAction` that is passed in and applies some default props to it.

packages/form/src/Field/Field.tsx
```diff
@@ -0,0 +1,197 @@
...
+import { COLOR } from "./Field.constants";
+
+type RootProps = Omit<
+  React.HTMLProps<HTMLInputElement>,
+  "value" | "onChange" | "size"
+> & {
+  disabled?: boolean;
+  error?: boolean | string;
+  value?: string;
+  onChange?: (
```
thomaswhyyou, line 16: Hm, this is surprising to see — why does this component need to care about `value` and `onChange`? Shouldn't it be the input component's concern?

packages/form/src/Field/Field.tsx
```diff
@@ -0,0 +1,197 @@
+import { Slot } from "@radix-ui/react-slot";
+import { TextField } from "@telegraph/text-field";
+import { Text as TelegraphText } from "@telegraph/typography";
+import { clsx } from "clsx";
+import React from "react";
+
+import { COLOR } from "./Field.constants";
+
+type RootProps = Omit<
```
thomaswhyyou, line 9: I'm just thinking out loud, but i'm guessing this component is meant to replace the internal implementation of `components/FormField/FormField.tsx`?

packages/text-field/src/TextField/TextField.tsx
```diff
@@ -0,0 +1,135 @@
...
+import { COLOR, SIZE } from "./TextField.constants";
+
+type TextFieldProps = Omit<
+  React.HTMLAttributes<HTMLInputElement>,
+  "value" | "onChange"
+> & {
+  size?: "1" | "2" | "3";
+  variant?: "outline" | "ghost";
+  disabled?: boolean;
+  error?: boolean | string;
```
thomaswhyyou, line 15: Does this component need to care about an error message? i.e. string value?

packages/text-field/src/TextField/TextField.tsx
```diff
@@ -0,0 +1,135 @@
...
+      disabled,
+      error,
+      ...props
+    },
+    forwardedRef,
+  ) => {
+    const inputRef = React.useRef<HTMLInputElement>(null);
+    const composedRefs = useComposedRefs(inputRef, forwardedRef);
+
+    const [internalState, setInternalState] =
```
thomaswhyyou, line 54: Same question as to why this needs to be an internal state.

packages/text-field/src/TextField/TextField.tsx
```diff
@@ -0,0 +1,135 @@
...
+          }}
+          className={clsx(
+            "appearance-none text-gray-12 border-none shadow-0 outline-0 bg-transparent",
+            "[font-family:inherit] h-full w-full",
+            SIZE.Field[size],
+            className,
+          )}
+          disabled={disabled}
+          {...props}
+          ref={composedRefs}
```
thomaswhyyou, line 116: I'm assuming this component is meant to be <input type="text" /> given the name. If so should the type be set explicitly? If not, should this be a generic `Input` component then?

packages/text-field/src/TextField/TextField.tsx
```diff
@@ -0,0 +1,135 @@
...
+> & {
+  size?: "1" | "2" | "3";
+  variant?: "outline" | "ghost";
+  disabled?: boolean;
+  error?: boolean | string;
+  value?: string;
+  onChange?: (
+    value: string,
+    { event }: { event: React.ChangeEvent<HTMLInputElement> },
+  ) => void;
```
thomaswhyyou, line 20: Hm, this is different than typical input/form handlers with an event as the first arg. Any reason for the divergence here?

packages/text-field/src/TextField/TextField.tsx
```diff
@@ -0,0 +1,135 @@
+import { useComposedRefs } from "@telegraph/compose-refs";
+import { Icon } from "@telegraph/icon";
+import { clsx } from "clsx";
+import React from "react";
+
```
thomaswhyyou, line 5: Personally find this component name confusing vs `Field`. i.e. This component does not render a label or a message if i'm understanding it correctly. 

Feels like it should be called `TextInput`?

thomaswhyyou (commented):
Haven't finished looking in details, but submitting my initial notes given I had enough high level questions/comments from the first pass.

I haven't seen the designs for the env vars project, but I wonder if the move here is to 1) limit the telegraph component to a low level input component and 2) just use `components/FormField/FormField.tsx` for the time being. Then a bigger effort later to replace `FormField.tsx` with a telegraph one. Happy to chat, curious what others think.

examples/playground/app/globals.css
```diff
@@ -1,7 +1,9 @@
-@import "@telegraph/tokens";
+@import "@telegraph/tokens/light";
```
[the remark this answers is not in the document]
kylemcd: This is in the playground, was just testing light mode 👍

examples/playground/app/page.tsx
```diff
@@ -1,11 +1,54 @@
 "use client";
 import { Heading } from "@telegraph/typography";
 import { Button } from "@telegraph/button";
-import { Icon, addSharp, chevronDown } from "@telegraph/icon";
+import { Icon, searchOutline, chevronDown, addSharp } from "@telegraph/icon";
+import { Field } from "@telegraph/form";
+import { TextField } from "@telegraph/text-field";
```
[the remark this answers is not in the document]
kylemcd: `Field` wraps any input and includes the label, message, and control components, it also controls all of the logic for handling error states. `TextField` is the individual `<input/>` that we styled to telegraph spec, this is separate so we can have an escape hatch to use this without needing to include `Field`. 

The naming might be confusing here, so open for suggestion on naming.

packages/compose-refs/src/compose-refs/compose-refs.ts
```diff
@@ -0,0 +1,35 @@
+// From Radix Primitives: https://github.com/radix-ui/primitives/blob/main/packages/react/compose-refs/src/composeRefs.tsx
```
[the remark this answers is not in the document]
kylemcd: You can see in the link in that file that the package is for internal use in radix only, so it would be unsafe to install this directly as it can contain breaking changes on any update. 

For each package the only file at the root of `src` is `index.ts`, anything else is contained in a folder. So being consistent with that logic here.

packages/compose-refs/src/index.ts
```diff
@@ -0,0 +1 @@
+export { composeRefs, useComposedRefs } from "./compose-refs";
```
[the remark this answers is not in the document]
kylemcd: For each package the only file at the root of src is index.ts, anything else is contained in a folder. So being consistent with that logic here.

packages/form/src/Field/Field.tsx
```diff
@@ -0,0 +1,197 @@
...
+  state: "default",
+  message: "",
+  value: "",
+  onChange: () => {},
+});
+
+const Root = React.forwardRef<RootRef, RootProps>(
+  ({ value, onChange, disabled, error, className, ...props }, forwardedRef) => {
+    const id = React.useId();
+    const [internalState, setInternalState] =
```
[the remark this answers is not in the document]
kylemcd: We'd only want 1 state to show at a time, with disabled overriding "default" or "error".  So this makes a mini state machine that allows for only one state at time so we don't end up in an "error" and "disabled" state at the same time. This couldn't be accomplished without internal state.

packages/form/src/Field/Field.tsx
```diff
@@ -0,0 +1,197 @@
...
+      React.useState<InternalProps["state"]>("default");
```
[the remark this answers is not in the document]
kylemcd: See my previous comment, disabled should override error.

packages/form/src/Field/Field.tsx
```diff
@@ -0,0 +1,197 @@
...
+    const [message, setMessage] = React.useState("");
+
+    React.useEffect(() => {
+      setMessage("");
```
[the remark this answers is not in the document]
kylemcd: `undefined` isn't assignable to the message state value as it's a string, so an empty string works here as it's falsy.

packages/form/src/Field/Field.tsx
```diff
@@ -0,0 +1,197 @@
...
+  error,
+  ...props
+}: DefaultProps) => {
+  return (
+    <Root value={value} onChange={onChange} disabled={disabled} error={error}>
+      {label && <Label>{label}</Label>}
+      {control && (
+        <Control>
+          {control.name === "TextField" && (
+            <TextField {...props} {...control.props} />
```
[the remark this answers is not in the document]
kylemcd: Control is that, and you can call `Field.Control` to do that. In the other telegraph components the `Default` component boilerplates a common use case so you don't have to compose the individual parts.

packages/form/src/Field/Field.tsx
kylemcd: This is needed as it stores `value` and `onChange` in context so it can be used in the `Control` component.

packages/form/src/Field/Field.tsx
kylemcd: Not understanding

packages/text-field/src/TextField/TextField.tsx
kylemcd: Yes, the border of the TextField turns red when there is an error.

packages/text-field/src/TextField/TextField.tsx
kylemcd: See other comment about internal state.

packages/text-field/src/TextField/TextField.tsx
kylemcd: I'm not understanding

packages/text-field/src/TextField/TextField.tsx
kylemcd: 99% of the time all you want is `event.target.value` so this makes it easier to grab that. It also helps for when we add a custom Combobox to `Field` where there won't be a ChangeEvent, so then there is just a similar value that can be used across both.

packages/text-field/src/TextField/TextField.tsx
kylemcd: Can change the name, was just following the naming convention in Figma.

thomaswhyyou:
Coming back to this, will share more thoughts this morning.

thomaswhyyou (commented):
Okay, have gone through this in more details, and shared additional thoughts. Let me know if it'd be helpful to chat through this. 

My biggest question mark right now is around `Field`. I almost want to suggest 1) we separate that out from this PR for a future consideration 2) we use the `TextField` (or rename to `Input`) with `FormField.tsx` for the first cut 3) then come back to figure out how we want to go about `Field` vs `FormField`.

packages/text-field/src/TextField/TextField.tsx
kylemcd: `error` is being passed from `Field` which is a `string` or `boolean` value. `string` when there is an error message `boolean` when there is no message but an error state is triggered. Thus, it would be incorrectly typed in `TextField` if we constrained this to just a boolean when it can be a string.

packages/text-field/src/TextField/TextField.tsx
kylemcd: I didn't add `type="text"` to this because things like `password` and `email` exist which would work fine here. I could narrow the type prop to only accept those types?

packages/text-field/src/TextField/TextField.tsx
kylemcd: I'm just concerned with handling non-native inputs like a combobox where there is no event being returned, this provides some consistency. But you're probably right that this shouldn't be happening in `TextField` but maybe rather `Field`.

packages/text-field/src/TextField/TextField.tsx
kylemcd: I strongly disagree with that statement. I think the right answer here is to change the name and make it consistent across both code and figma. But deviating on naming is going to lead to confusion later.

packages/text-field/src/TextField/TextField.tsx
```diff
@@ -0,0 +1,135 @@
...
+          onChange={(event) => {
+            onChange?.(event.target.value, { event });
+          }}
+          className={clsx(
+            "appearance-none text-gray-12 border-none shadow-0 outline-0 bg-transparent",
+            "[font-family:inherit] h-full w-full",
+            SIZE.Field[size],
+            className,
+          )}
+          disabled={disabled}
```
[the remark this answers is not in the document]
kylemcd: We destructure it earlier in the file, so no it's not included in props.

packages/text-field/src/TextField/TextField.tsx
```diff
@@ -0,0 +1,135 @@
...
+  variant?: "outline" | "ghost";
+  disabled?: boolean;
+  error?: boolean | string;
+  value?: string;
+  onChange?: (
+    value: string,
+    { event }: { event: React.ChangeEvent<HTMLInputElement> },
+  ) => void;
+  leadingIcon?: React.ComponentProps<typeof Icon>;
+  TrailingAction?: React.ReactElement;
```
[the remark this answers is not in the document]
kylemcd: Following the direction of the design system here. We could make them take a react element for both but it would be an early abstraction for functionality that's not needed. 

The correct name should be `TrailingAction` as it can accept more than a button, a tooltip for example.

packages/form/src/Field/Field.tsx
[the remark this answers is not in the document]
kylemcd: We pass `internalState` to the context provider value `state`, this is then used in the other components in this file. 

Are you proposing something more like this?
```
const internalState = disabled ? "disabled" : error ? "error" : "default"
React.useEffect(() => {
   setMessage("")
   if(internalState === "error" && typeof error === "string"){
        setMessage(error)
   }
} , [internalState])
```

packages/form/src/Field/Field.tsx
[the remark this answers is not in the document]
kylemcd: The alternative being something like this where you need to import and include the TextField each time vs just changing a prop on `Field`? I'd prefer this but was met with opposition to a composition pattern like this before which is why I'm creating these `Default` components as shortcuts.
```
<Field label="label">
  <TextField />
</Field>
```

packages/text-field/src/TextField/TextField.tsx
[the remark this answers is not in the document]
thomaswhyyou: > The correct name should be TrailingAction as it can accept more than a button, a tooltip for example.

Ah yep got it.

> We could make them take a react element for both but it would be an early abstraction for functionality that's not needed.

My 2c is the current prop definitions for these two feel perhaps overly specific. Personally it'd be nice to simplify/unify as a start. But not a blocker if you feel strongly about it!

packages/text-field/src/TextField/TextField.tsx
thomaswhyyou: Maybe the move is to just make this an `Input` component then, it seems that's what it is. I know that is not a thing in the figma file, but then again the implmementation and the figma design system should evolve together to match.

packages/form/src/Field/Field.tsx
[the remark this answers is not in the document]
thomaswhyyou: I think we probably have a different mental modal on this. For me `TextField` is and should be a standalone component that you can use with or without `Field`. So in my mind `Field` is not necessarily a top level component that owns lower level components like `TextField` and therefore should provide a default. They are hierarchical in the markup structure, but conceptually standalone.

packages/form/src/Field/Field.tsx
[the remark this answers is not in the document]
thomaswhyyou: No without the internal state, like:
```
const state = disabled ? "disabled" : error ? "error" : "default"

return (
      <FieldContext.Provider
        value={{ state, id, message, value, onChange }}
      >
)
```

Maybe i'm being completely stupid, but I don't follow why we need an internal state in both components 😬

thomaswhyyou (commented):
Responded to a few, but i think we might need to chat through some of the broader points. In the mean time I'm curious to hear others' opinions also. Sorry this is dragging a bit!

PR #61, opened 2024-03-05
feat: add storybook

kylemcd:
- Adds storybook so we can view, qa, build, and test our components
- Adds `stories` file for each existing component


packages/typography/src/Text/Test.stories.tsx
cjbell, line 1: Nit: file should be `Text.stories.tsx` not `Test`

cjbell (approved):
one nit but looks great otherwise 💯 ✨

packages/typography/src/Text/Test.stories.tsx
kylemcd: Nice catch

PR #64, opened 2024-03-06
feat: as prop on button

kylemcd:
`@telegraph/button`
- Adds `as` prop to Button so that we can pass other components to be used in place of a standard "button". This also uses a typescript generic so that we can cast the passed in `as` component's props onto the button. For example, when passing a Next.js Link through the `as` prop, you'll get `href`, `shallow`,`scroll`, etc, props automatically.

### DIFF
--- a/packages/button/src/Button/Button.tsx
+++ b/packages/button/src/Button/Button.tsx
@@ -13,6 +13,25 @@ import {
   textSizeMap,
 } from "./Button.constants";
 
+
+// The `as` prop is a generic prop that allows you to change the underlying
+// element of a component. This is useful for creating a component that can
+// be used as a button, link, or any other element.
+type AsProp<C extends React.ElementType> = {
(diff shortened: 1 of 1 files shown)


packages/button/src/Button/Button.tsx
```diff
@@ -13,6 +13,24 @@ import {
...
 } from "./Button.constants";
 
+// The `as` prop is a generic prop that allows you to change the underlying
+// element of a component. This is useful for creating a component that can
+// be used as a button, link, or any other element.
+type AsProp<C extends React.ElementType> = {
+  as?: C;
+};
+
+// The `PropsWithAs` type is a utility type that allows you to create a
```
connorlindsey, line 23: Thanks for the detailed comments here. Complex typescript can get gnarly

connorlindsey (approved):

PR #65, opened 2024-03-07
feat: input component

kylemcd:
Environment for visual QA: https://telegraph-storybook-git-kyle-kno-5399-telegrap-63efba-knocklabs.vercel.app/?path=%2Fstory%2Fcomponents-input--default

`@telegraph/input`
- Adds input component
- Adds associated storybook story

`@telegraph/compose-refs`
- Adds ref composing utility used internally in Radix

`@telegraph/button`
- Modify data attributes for targeting
- Update failing test

`@telegraph/icon`
- Modify data attributes for targeting


packages/button/src/Button/Button.tsx
```diff
@@ -84,7 +84,8 @@ const Root = React.forwardRef<RootRef, RootProps>(
             buttonSizeMap[layout][size],
             className,
           )}
-          data-button-layout={layout}
+          data-tgph-button
+          data-tgph-button-layout={layout}
```
kylemcd, line 88: Used for granular css targeting. Also added `tgph`prefix to avoid future collisions.

packages/compose-refs/src/compose-refs.ts
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * A custom hook that composes multiple refs
+ * Accepts callback refs and RefObject(s)
+ */
+function useComposedRefs<T>(...refs: PossibleRef<T>[]) {
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  return React.useCallback(composeRefs(...refs), refs);
+}
+
+export { composeRefs, useComposedRefs };
```
kylemcd, line 35: Taken from internal utility that Radix uses. Would be unsafe to install directly as it's only used internally, so brought it into our project here instead.

packages/input/src/Input/Input.tsx
```diff
@@ -0,0 +1,157 @@
...
+      error,
+      children,
+      ...props
+    },
+    forwardedRef,
+  ) => {
+    const inputRef = React.useRef<HTMLInputElement>(null);
+    const composedRefs = useComposedRefs(forwardedRef, inputRef);
+
+    const state = disabled ? "disabled" : error ? "error" : "default";
```
kylemcd, line 49: Derives state for displaying the right visuals on the input. Disabled is then passed into the input to get out of the box browser functionality.

packages/input/src/Input/Input.tsx
```diff
@@ -0,0 +1,157 @@
...
+              return;
+            }
+
+            const input = inputRef.current;
+            if (!input) return;
+
+            requestAnimationFrame(() => {
+              input.focus();
+            });
+          }}
```
kylemcd, line 76: Does 2 things:
- Doesn't focus input when clicking on a button inside of the input
- Focuses input when clicking anywhere in the container, for example on the leading icon that isn't a button.

packages/input/src/Input/Input.tsx
```diff
@@ -0,0 +1,157 @@
...
+          "group box-border flex items-center justify-center h-full",
+          "[&>[data-tgph-button]]:w-full [&>[data-tgph-button]]:h-auto",
+          // Overrides to match the icon button in figma
+          "[&>[data-tgph-button-layout='icon-only']]:aspect-square [&>[data-tgph-button-layout='icon-only']]:p-0",
+          // Overrides default button layout to match the button in figma
+          "[&>[data-tgph-button-layout='default']]:px-2",
+          // If only an icon button is present, set the aspect ratio to square
+          "[&:has([data-tgph-button-layout='icon-only'])]:aspect-square",
+          // If only an icon button is present, reset the padding to spacing-1
+          "[&:has([data-tgph-button-layout='icon-only'])]:!p-1",
```
kylemcd, line 118: Overrides some styles for the Button `icon-only` and `default` layouts so it fits into each size of the text field like figma outlines.

packages/input/src/Input/Input.tsx
```diff
@@ -0,0 +1,157 @@
...
+          position === "leading" && SIZE.SlotLeading[context.size],
+          position === "trailing" && SIZE.SlotTrailing[context.size],
+          SIZE.Slot[props.size ?? context.size],
+        )}
+      >
+        <RadixSlot size={context.size} {...props} ref={forwardedRef} />
+      </div>
+    );
+  },
+);
```
kylemcd, line 128: Component used to pass "any" component into the leading or trailing position of the form field.

packages/input/src/Input/Input.tsx
```diff
@@ -0,0 +1,157 @@
...
+}: DefaultProps) => {
+  return (
+    <Root {...props}>
+      {LeadingComponent && <Slot position="leading">{LeadingComponent}</Slot>}
+      {TrailingComponent && (
+        <Slot position="trailing">{TrailingComponent}</Slot>
+      )}
+    </Root>
+  );
+};
```
kylemcd, line 148: This component is only for making composition of the input quicker like this:
```
<Input size="3" LeadingComponent={<Button} />
```

You can always drop down to the individual parts and build an input like this:

Instead of utilizing the underlying components like this
```
<Input.Root>
    <Input.Slot position="leading">
        <Button/>
     </Input.Slot>
</Input.Root>

File: packages/input/src/Input/Input.tsx [hunk 8]
```diff
@@ -0,0 +1,157 @@
...
+const Slot = React.forwardRef<SlotRef, SlotProps>(
+  ({ position = "leading", ...props }, forwardedRef) => {
+    const context = React.useContext(InputContext);
+    return (
+      <div
+        className={clsx(
+          "group box-border flex items-center justify-center h-full",
+          "[&>[data-tgph-button]]:w-full [&>[data-tgph-button]]:h-auto",
+          // Overrides to match the icon button in figma
+          "[&>[data-tgph-button-layout='icon-only']]:aspect-square [&>[data-tgph-button-layout='icon-only']]:p-0",
```
cjbell, line 112: Would it be better here to pull these out into its own CSS file? I know end result is similar, just thinking about maintainability and some of these tailwind classes get hard to read

packages/input/src/Input/Input.tsx
```diff
@@ -0,0 +1,157 @@
...
+import { useComposedRefs } from "@telegraph/compose-refs";
+import clsx from "clsx";
+import React from "react";
+
+import { COLOR, SIZE } from "./Input.constants";
+
+type BaseRootProps = {
+  size?: keyof typeof SIZE.Container;
+  variant?: keyof typeof COLOR.Container.default;
+  error?: boolean;
```
cjbell, line 11: Generally would prefer “hasError” or “errored” for this given it take a boolean

cjbell (approved):
This is great, Kyle! 2 comments but not blocking!

packages/input/src/Input/Input.tsx
kylemcd: Eventually for the wrapping "form control" element I was planning on having that take an `error` prop that could be a `string` or `boolean`. `string` in the case of wanting to display an error message + change colors to reflect an error state and `boolean` to change colors to accept an error state. 

Do you think it makes sense knowing that plan to make this `hasError` and leave the other as `error` or to keep them both consistent as `error`? 

Fine with either option, just want to make sure as we create more components with error states we follow the same pattern throughout.

packages/input/src/Input/Input.tsx
kylemcd: With scoping and how that's wired through tailwind I'd like to keep all of these through tailwind as much as possible so that it's one less vector for something to go wrong later on. I think I'd be fine with the approach of having the css file if all of the styles were through that, but to mix and match just seems like it could get muddy later on.

packages/input/src/Input/Input.tsx
cjbell: To me I think they should be different and the input should be “hasError” that the field component sets derived from the error string. Keeps a nice clean API on the input.

kylemcd: Cool, works for me 👍

packages/input/src/Input/Input.tsx
```diff
@@ -0,0 +1,157 @@
...
+);
+
+type DefaultProps = React.ComponentPropsWithoutRef<typeof Root> & {
+  LeadingComponent?: React.ReactNode;
+  TrailingComponent?: React.ReactNode;
+};
+
+const Default = ({
+  LeadingComponent,
+  TrailingComponent,
```
thomaswhyyou, line 137: Nit, but these capitalized prop names kinda throw me off because based on the code example given below i think they are instantiated component elements, not the component functions?
```
// we are supposed to do this
<Input size="3" LeadingComponent={<Button />} />

// but makes me think we are supposed to use it like this
<Input size="3" LeadingComponent={Button} />
```

Feel free to ignore if you feel strongly about this, but I would personally want to see non-capitalized names (and maybe something shorter) i.e. `leading` or `leadingComponent`.

packages/input/src/Input/Input.tsx
```diff
@@ -0,0 +1,157 @@
...
+import { useComposedRefs } from "@telegraph/compose-refs";
+import clsx from "clsx";
+import React from "react";
+
+import { COLOR, SIZE } from "./Input.constants";
+
+type BaseRootProps = {
+  size?: keyof typeof SIZE.Container;
+  variant?: keyof typeof COLOR.Container.default;
+  hasError?: boolean;
```
thomaswhyyou, line 11: I know this is an ultra naming nit 🫣 and feel free to ignore - but slight preference toward `errored` or `isErrored`, to be in line with `disabled`. Maybe `isErrored` better to distinguish the native input attribute vs the component one.

thomaswhyyou (approved):
Makes sense to me, and thanks for breaking this one out!

PR #66, opened 2024-03-07
feat: box component 

kylemcd:
`@telegraph/layout`
- Adds the base box component


packages/layout/package.json
```diff
@@ -0,0 +1,60 @@
+{
+  "name": "@telegraph/layout",
```
kylemcd, line 2: Moved what was `@telegraph/box` to `@telegraph/layout` to simplify importing these layouts. Also matches chakra's package naming scheme for similar components.

packages/layout/src/Box/Box.constants.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export const BOX_PROPS = {
+  p: {
+    rule: "padding",
+    type: "spacing",
+  },
+  m: {
+    rule: "margin",
+    type: "spacing",
+  },
```
kylemcd, line 13: Notice that this is now an object. This makes it so we can access each box prop by object key instead of needing to loop through this each time.

packages/layout/src/Box/Box.tsx
```diff
@@ -0,0 +1,89 @@
...
+};
+type BoxRef = HTMLDivElement;
+
+const Box = React.forwardRef<BoxRef, BoxProps>(
+  ({ variant = "ghost", className, ...props }) => {
+    // TODO: compose this with forwardRef once @telegraph/composed-ref is published
+    const boxRef = React.useRef<HTMLDivElement>(null);
+
+    // Filter out the box props from the rest of the props
+    const filteredProps = Object.keys(props).reduce(
```
cjbell, line 24: I don’t think this is strictly necessary, but given how stable these props are is it worth memoizing this? Thinking about how often we’re going to be using these components and how many will exist in the tree it might be worth it?

packages/layout/package.json
cjbell: Yeah I like this naming a lot!

cjbell (approved):
Looks great and love the naming and conventions!

PR #68, opened 2024-03-11
feat: stack component + responsive props

kylemcd:
`@telegraph/layout`
- Adds Stack component
- Adds responsive props to layout component values
- Modularize and co-locate helper functions

`@telegraph/postcss-config`
- Adds custom media query plugin


examples/playground/app/page.tsx
```diff
@@ -3,14 +3,25 @@ import { Heading } from "@telegraph/typography";
...
-import { Box } from "@telegraph/layout";
+import { Box, Stack } from "@telegraph/layout";
 
 export default function Home() {
   return (
     <main className="tgph">
-      <Box pl="10" pt="40">
-        I am a box
-      </Box>
+      <Box pl={{ sm: "0", lg: "40" }}>I am a box</Box>
```
kylemcd, line 11: Example of responsive values in action.

packages/layout/src/Box/Box.styles.css
```diff
@@ -1,17 +1,38 @@
...
+    --tgph-padding: var(--tgph-padding-xl);
+    --tgph-margin: var(--tgph-margin-xl);
+  }
+}
+
+@media (--tgph-breakpoint-xxl) {
+  .tgph-box {
+    --tgph-padding: var(--tgph-padding-xxl);
+    --tgph-margin: var(--tgph-margin-xxl);
+  }
```
kylemcd, line 37: This sets the css variable to the correct value based on the screen size.

packages/layout/src/Box/Box.tsx
```diff
@@ -41,49 +44,17 @@ const Box = React.forwardRef<BoxRef, BoxProps>(
...
-            boxRef.current.style.setProperty(key, value);
-          }
-        });
-      };
-      setCssVariables();
+      propsToCssVariables({
+        props: filteredProps.box,
+        ref: boxRef,
+        propMap: BOX_PROPS,
+      });
```
kylemcd, line 51: This functionality is moved into the helpers folder.

packages/layout/src/helpers/breakpoints.ts
```diff
@@ -0,0 +1,52 @@
+export type ResponsiveObject<T> = { sm?: T; md?: T; lg?: T; xl?: T; xxl?: T };
+export type Responsive<T> = T | ResponsiveObject<T>;
+
+export const BREAKPOINTS = ["sm", "md", "lg", "xl", "xxl"];
+export const getValueForEachBreakpoint = (
+  breakpoints: ResponsiveObject<string>,
+) => {
```
kylemcd, line 7: This function gets values for each breakpoint based on the object passed in. This also fills out values for breakpoints that were not passed in. 

For example
```
{ md: "20", xxl: "2" }
```

If this object is passed in sm-md are set to `20`, and lg-xxl are set to `2`. This works the same way as min width media queries work in css.

packages/tokens/scripts/generate-css-file.js
```diff
@@ -69,6 +69,10 @@ const tokensToCss = (variables = {}, mode = "tokens") => {
           } else {
             regularVars += `  --${TELEGRAPH_VARIABLE_PREFIX}${currentPrefix}: ${value};`;
           }
+        } else if (mode === "breakpoint") {
+          if (currentPrefix.includes("breakpoint")) {
+            regularVars += `  @custom-media --${TELEGRAPH_VARIABLE_PREFIX}${currentPrefix} (min-width: ${value});`;
+          }
```
kylemcd, line 75: Creates a `breakpoint.css` file that we then use in the `postcss-config` in order to be able to utilize css variable syntax with `@custom-media`.

packages/tokens/src/tokens/breakpoint.ts
```diff
@@ -0,0 +1,7 @@
+export const breakpoint = {
+  sm: "600px",
+  md: "768px",
+  lg: "1024px",
+  xl: "1280px",
+  xxl: "1920px",
```
kylemcd, line 6: Would it make more sense to make these something like "phone" "tablet" "desktop" or the like?

packages/layout/src/helpers/breakpoints.ts
```diff
@@ -0,0 +1,52 @@
+export type ResponsiveObject<T> = { sm?: T; md?: T; lg?: T; xl?: T; xxl?: T };
```
thomaswhyyou, line 1: Looking at some of these helper functions made me wonder if there's any merit to writing some simple tests to verify / strap them in? Not suggesting we must btw, more of a thought that occurred to me as i'm looking over, and whether they would be helpful to us and for the future - especially since i don't necessarily grok everything they are doing from a glance.

thomaswhyyou (commented):
Left a few comments/questions but not a ton of notes though, I assume they are pretty straight forward utility components that should do what you think they will do 😄

packages/tokens/src/tokens/breakpoint.ts
cjbell: Plus one here to the sizes you have here and to using `2xl` over `xxl`!

packages/tokens/src/tokens/breakpoint.ts
```diff
@@ -0,0 +1,7 @@
+export const breakpoint = {
+  sm: "600px",
```
thomaswhyyou, line 2: One thing I meant to ask: where are these breakpoints coming from? Did we grab them from another framework defaults? Or is this our own definition?

packages/tokens/src/tokens/breakpoint.ts
kylemcd: Cool this works!

packages/layout/src/helpers/css-variables.ts
```diff
@@ -0,0 +1,83 @@
...
+    });
+  }
+
+  return value;
+};
+
+type PropToCssVariablesArgs = {
+  props: Record<string, Responsive<string>>;
+  ref: React.RefObject<HTMLElement>;
+  propsMap: Record<string, CssVariableProp>;
```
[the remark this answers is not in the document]
kylemcd: Yeah, what do you think would be a more straightforward name? `props` are the actual props passed into the component where as `propsMap` is the mapping of prop info in the constants file.

packages/tokens/src/tokens/breakpoint.ts
kylemcd: Didn't think to pull from a framework. I've just used these breakpoints since the dawn of time. Maybe it makes sense to grab the ones defined here by tailwind instead? https://tailwindcss.com/docs/screens

packages/layout/src/helpers/breakpoints.ts
kylemcd: Yeah I was going to write tests for all of these once we get functionality finalized!

packages/layout/src/Stack/Stack.tsx
```diff
@@ -0,0 +1,66 @@
...
+import React from "react";
+
+import { Box } from "../Box";
+import { type Responsive } from "../helpers/breakpoints";
+import { propsToCssVariables } from "../helpers/css-variables";
+
+import { STACK_PROPS } from "./Stack.constants";
+
+type StackProps = React.ComponentPropsWithoutRef<typeof Box> & {
+  gap: Responsive<`${keyof typeof t.tokens.spacing}`>;
```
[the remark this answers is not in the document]
kylemcd: It's meant to be optional with a default of zero, missed the `?` when I renamed it.

packages/tokens/scripts/generate-css-file.js
kylemcd: Yeah I can change the css file name. The idea for the tokens themselves is that they're singular, so you can reference `breakpoint-sm` instead of `breakpoints-sm`

packages/layout/src/Box/Box.tsx
```diff
@@ -1,13 +1,16 @@
...
 import { BOX_PROPS, VARIANT } from "./Box.constants";
-import { deriveSpacing } from "./Box.helpers";
 
 type BoxProp = keyof typeof BOX_PROPS;
 type BoxProps = React.HTMLAttributes<HTMLDivElement> & {
-  [key in BoxProp]?: `${keyof typeof t.tokens.spacing}`;
+  [key in BoxProp]?: Responsive<`${keyof typeof t.tokens.spacing}`>;
 } & {
   // More variants wil beed added here once
   // they are designed
```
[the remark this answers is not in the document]
thomaswhyyou: > like a box with a gray rounded border, would be added as variants here.

Hmm, that sounds kinda strange to me for Box, and maybe worth a chat at some point. But not blocking this PR.

packages/tokens/src/tokens/breakpoint.ts
thomaswhyyou: > Maybe it makes sense to grab the ones defined here by tailwind instead? https://tailwindcss.com/docs/screens

Yeah that seems like a pretty sensible choice / starting place? And document where that came from for completeness.

packages/layout/src/Stack/Stack.tsx
```diff
@@ -0,0 +1,66 @@
...
+};
+
+type StackRef = React.ElementRef<typeof Box>;
+
+const Stack = React.forwardRef<StackRef, StackProps>(
+  ({ className, ...props }, forwardedRef) => {
+    const stackRef = React.useRef<HTMLDivElement>(null);
+    const composedRef = useComposedRefs(forwardedRef, stackRef);
+
+    // Filter out the stac props from the rest of the props
```
thomaswhyyou, line 28:
```suggestion
    // Filter out the stack props from the rest of the props
```

packages/layout/src/Box/Box.tsx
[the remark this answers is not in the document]
kylemcd: Sounds good!

packages/layout/src/helpers/css-variables.ts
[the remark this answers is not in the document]
thomaswhyyou: Ah I seeee. Maybe `props` and `mappings`? Or maybe i'm overthinking here, because I initially thought the "map" in "propsMap" was referring to the data type i.e. Record, but now I see it's referring to the mapping. 

The name of this arg isn't super consequential at the end of the day, so I will defer the final call to you. I think fine to leave it be if we can't think of a better name.

packages/tokens/src/tokens/breakpoint.ts
```diff
@@ -0,0 +1,7 @@
+export const breakpoint = {
```
thomaswhyyou, line 1: Worth leaving a note here to say these are referenced from tailwind defaults?

thomaswhyyou (approved):
👏

PR #86, opened 2024-05-06
feat: add additional props to layout components

kylemcd:
- Adds height, width, border radius, and border props to the layout components
- Simplifies location based css values function 
- Creates type helpers for layout props
- Adds flattened tokens for use with layout components


packages/layout/src/Box/Box.constants.ts
```diff
@@ -1,62 +1,166 @@
...
     rule: "margin-block",
-    type: "spacing-y",
+    type: "spacing",
+    direction: "y",
+  },
+  bg: {
+    rule: "background-color",
+    type: "color",
+  },
+  r: {
```
connorlindsey, line 84: What're your thoughts on using shorthand properties like `r` vs. more verbose ones like `rounded` (same for `bd` vs. `border`)? 

I know we use `m` / `p` which are common enough that they're easy to understand but `r` and `bd` are a bit harder to parse, especially when you start adding modifiers like direction and additional properties (e.g. `bdc`). 

I'm not strongly opposed to the shorter properties as they're nicer to write, but I'm also worried it'll make components harder to read and the component api harder to learn/remember.

kylemcd: Yeah I was going back and forth myself on it. I think between your thought and sleeping on it verbosity makes more sense for these. Easy swap 👍

connorlindsey (approved):
🎉

kylemcd:
* **#87** <a href="https://app.graphite.dev/github/pr/knocklabs/telegraph/87?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#86** <a href="https://app.graphite.dev/github/pr/knocklabs/telegraph/86?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #94, opened 2024-05-30
fix: button whitespace wrapping

connorlindsey:
### Before
Screenshot 2024-05-30 at 12 00 06 PM

### After
Screenshot 2024-05-30 at 12 00 16 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "node": "18.x"
   },
   "scripts": {
-    "dev:packages": "turbo dev --filter=\"./packages/*\"",
+    "dev:packages": "turbo dev --filter=\"./packages/*\" --concurrency=100%",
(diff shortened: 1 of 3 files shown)


kylemcd (approved):

package.json
```diff
@@ -12,7 +12,7 @@
     "node": "18.x"
   },
   "scripts": {
-    "dev:packages": "turbo dev --filter=\"./packages/*\"",
+    "dev:packages": "turbo dev --filter=\"./packages/*\" --concurrency=100%",
```
connorlindsey, line 15: I hit Turbo's default concurrency limit while running storybook + dev:packages which I think is a result of how many packages we have https://turbo.build/repo/docs/reference/command-line-reference/run#--concurrency

PR #98, opened 2024-06-04
feat: modal

kylemcd:
### Description
- Fix modal positioning issues
- Add Body, Heading, and Footer components
- Add a11y title and description
- Add storybook story


kylemcd:
* **#98** <a href="https://app.graphite.dev/github/pr/knocklabs/telegraph/98?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

packages/modal/src/Modal/Modal.stories.tsx
```diff
@@ -0,0 +1,31 @@
+import type { Meta } from "@storybook/react";
+import { Button } from "@telegraph/button";
+import { Heading } from "@telegraph/typography";
+
+import { Modal as TelegraphModal } from "./Modal";
+
+const meta: Meta<typeof TelegraphModal> = {
+  title: "Components/Modal",
+  // component: TelegraphModal,
```
connorlindsey, line 9: Should this be commented out?

packages/modal/src/Modal/Modal.stories.tsx
```diff
@@ -0,0 +1,31 @@
...
+};
+export default meta;
+
+export const Modal = {
+  render: () => (
+    <TelegraphModal.Root open={true}>
```
connorlindsey, line 15: Are you getting typesense for these props? Testing locally with the story, I'm only seeing `key?` and `ref?` as props. Not a blocking issue, but it'd be great to have that so these components are easier to work with in Control.

connorlindsey, line 15: Nit: Would be nice to have an open/close button to fully test out the modal and show the open/close animation.

connorlindsey (approved):
Looks awesome! Just a couple non-blocking comments

packages/modal/src/Modal/Modal.stories.tsx
kylemcd: Because it's not a "Default" component and rather the `Modal.ComponentPart` pattern it doesn't work great with storyboook. Once we decide what a good default it is we'll swap that in here and the type sense will be better.

PR #109, opened 2024-06-06
feat: icon alt prop optional when aria-hidden is present

kylemcd:
### Description
- If an icon is only used for decoration the `alt` prop should not be required. So, if passing `aria-hidden` to the Icon, the `alt` prop becomes optional as this is the correct prop to pass to adhere to a11y standards.

### Tasks
[KNO-6152](https://linear.app/knock/issue/KNO-6152/[telegraph]-icon-optional-alt-when-passing-aria-hidden)

### DIFF
--- a/.changeset/seven-clouds-applaud.md
+++ b/.changeset/seven-clouds-applaud.md
@@ -1,0 +1,5 @@
+---
+"@telegraph/icon": patch
+---
+
+When aria-hidden prop is preset on `<Icon/>`, alt prop is optional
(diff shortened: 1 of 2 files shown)


kylemcd:
* **#109** <a href="https://app.graphite.dev/github/pr/knocklabs/telegraph/109?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

connorlindsey (approved):

connorlindsey:
@kylemcd Will this also work for the icon props on the button component? Assuming so but want to double check

kylemcd:
> @kylemcd Will this also work for the icon props on the button component? Assuming so but want to double check

@connorlindsey Yep tested and it works there as well 👍

PR #118, opened 2024-06-10
fix: don't throw error if alt text is missing for aria hidden icon

connorlindsey:
Missing alt text would still throw a runtime error even if aria hidden is set to true. Also adds a test to check that behavior.


kylemcd (approved):
Oh dang, not sure how this didn't throw an error for me when doing this.

PR #135, opened 2024-06-20
feat: combobox component

kylemcd:
feat: combobox component

chore: format


kylemcd:
* **#135** <a href="https://app.graphite.dev/github/pr/knocklabs/telegraph/135?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

packages/button/src/Button/Button.constants.ts
```diff
@@ -59,14 +59,44 @@ export const buttonColorMap = {
...
+      h: "8",
+      gap: "0",
+      px: "0",
+    },
+    "3": {
+      w: "10",
+      h: "10",
+      gap: "0",
+      px: "0",
+    },
```
kylemcd, line 99: There were competing classes and Box attributes, we should just use box attributes so set these here.

packages/combobox/src/Combobox/Combobox.helpers.ts
```diff
@@ -0,0 +1,12 @@
...
+  value: Option | Array<Option>,
+): value is Array<Option> => {
+  return Array.isArray(value);
+};
+
+export const isSingleSelect = (
+  value: Option | Array<Option>,
+): value is Option => {
+  return typeof value === "object" && !Array.isArray(value);
+};
```
kylemcd, line 12: For narrowing types in the Combobox so we can properly type instances of Single Select & Multi Select

packages/combobox/src/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,740 @@
...
+                      }}
+                      exit={{
+                        opacity: 0.5,
+                      }}
+                      transition={{ duration: 0.1, type: "spring", bounce: 0 }}
+                      key={n}
+                    >
+                      {n}
+                    </Box>
+                  ))}
```
kylemcd, line 277: Split the characters of the numbers of "+3 more" to smoothly animate them in and out.

packages/combobox/src/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,740 @@
...
+        role="button"
+        icon={{ icon: Lucide.X, alt: `Remove ${value}` }}
+        onClick={(event: React.MouseEvent) => {
+          const onValueChange = context.onValueChange as (
+            v: Array<Option>,
+          ) => void;
+          const contextValue = context.value as Array<Option>;
+          const newValue = contextValue.filter((v) => v.value !== value);
+          onValueChange(newValue);
+          event.stopPropagation();
```
kylemcd, line 177: stopPropagation so that we don't open or close the combobox when removing a tag.

packages/combobox/src/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,740 @@
...
+    return () => observer.disconnect();
+  }, [initialAnimationComplete, setHeightFromContent]);
+
+  // Reset the animation complete state when the combobox is closed
+  React.useEffect(() => {
+    if (initialAnimationComplete === true && context.open === false) {
+      setInitialAnimationComplete(false);
+    }
+  }, [context.open, initialAnimationComplete]);
+
```
kylemcd, line 426: Logic for creating the smooth height animation change as you're searching inside of the combobox.

packages/layout/src/Box/Box.styles.css
```diff
@@ -1,8 +1,7 @@
 .tgph-box {
   display: var(--tgph-display, block);
   padding: var(--tgph-padding);
-  /* Tailwind css reset overrides this, so we need important specificity */
-  margin: var(--tgph-margin) !important;
+  margin: var(--tgph-margin);
```
kylemcd, line 4: Realized this was only happening because of a ghost package laying around in my local env.

connorlindsey (approved):
Love it!

packages/combobox/src/Combobox/Combobox.tsx
cjbell, line 1: One note here: would it be better to split up the sub components in this file to make this a bit easier to follow? Not sure on the conventions we've been following in tgraph to date, but feels like 800 lines in here is a fair bit!

cjbell (approved):
💯💯💯

packages/combobox/src/Combobox/Combobox.tsx
kylemcd: Yeah I was thinking about that. Generally I like to keep all related parts of a component in one file, separating it out just makes it harder to work with IMO. I think this component is teetering on the edge of where that complexity would benefit from being separated but I think for now it's fine.

PR #152, opened 2024-06-28
feat: introduce size 0 tag component

cjbell:
* Size 0 tag
* Sets different left padding
* Cannot have `onCopy` or `onRemove`

### DIFF
--- a/packages/icon/src/Icon/Icon.tsx
+++ b/packages/icon/src/Icon/Icon.tsx
@@ -64,7 +64,12 @@ const Icon = <T extends TgphElement>({
       {...props}
     >
       {IconComponent && (
-        <IconComponent aria-label={alt} width="100%" height="100%" />
+        <IconComponent
+          aria-label={alt}
+          width="100%"
+          height="100%"
+          display="block"
+        />
       )}
     </Box>
   );
--- a/packages/tag/src/Tag/Tag.tsx
+++ b/packages/tag/src/Tag/Tag.tsx
@@ -14,7 +14,7 @@ import { clsx } from "clsx";
 import { AnimatePresence, motion } from "framer-motion";
 import React from "react";
 
-import { COLOR, SIZE } from "./Tag.constants";
+import { COLOR, SIZE, SPACING } from "./Tag.constants";
 
 type RootBaseProps = {
   size?: "0" | "1" | "2";
@@ -50,7 +50,7 @@ const Root = <T extends TgphElement>({
         display="inline-flex"
         align="center"
         rounded="3"
-        pl={size === "0" ? "1" : "2"}
+        pl={SPACING.Root[size]}
         className={clsx(SIZE.Root[size], COLOR.Root[variant][color], className)}
         {...props}
         data-tag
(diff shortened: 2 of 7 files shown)


packages/icon/src/Icon/Icon.tsx
```diff
@@ -64,7 +64,12 @@ const Icon = <T extends TgphElement>({
       {...props}
     >
       {IconComponent && (
-        <IconComponent aria-label={alt} width="100%" height="100%" />
+        <IconComponent
+          aria-label={alt}
+          width="100%"
+          height="100%"
+          display="block"
```
cjbell, line 71: When the parent container for an icon is inline-block you always need this so the icon acts appropriately in the container

packages/tag/src/Tag/Tag.tsx
```diff
@@ -227,10 +228,12 @@ const Default = <T extends TgphElement>({
...
       {icon && <Icon {...icon} />}
       <Text as="span">{children}</Text>
-      {onRemove && (
+      {onRemove && size !== "0" && (
         <Button onClick={onRemove} icon={{ icon: Lucide.X, alt: "Remove" }} />
       )}
-      {onCopy && <CopyButton onClick={onCopy} textToCopy={textToCopy} />}
+      {onCopy && size !== "0" && (
+        <CopyButton onClick={onCopy} textToCopy={textToCopy} />
+      )}
```
kylemcd, line 236: I think gating this by `size` is an unnecessary abstraction for this.

packages/tag/src/Tag/Tag.tsx
```diff
@@ -76,6 +76,7 @@ const Text = <T extends TgphElement>({
       as={as}
       size={context.size}
       color={COLOR.Text[context.variant][context.color]}
+      weight="medium"
```
cjbell, line 79: Tags were default to regular, which doesn't match the design

packages/tag/src/Tag/Tag.tsx
```diff
@@ -50,7 +50,7 @@ const Root = <T extends TgphElement>({
         display="inline-flex"
         align="center"
         rounded="3"
-        pl="2"
+        pl={size === "0" ? "1" : "2"}
```
kylemcd, line 53: Can we add this to a map in Tags.constants instead? Ideally all of these are configuration based and not inline prop logic. 

```
const rootSpacing = {
"0": "1",
"1": "2",
....
"
```

packages/tag/src/Tag/Tag.tsx
cjbell: If we don't do that then we need to build a size 0 button fwiw

kylemcd: Yeah, I get that it won't work. But nothing is stopping you from doing this either. 
```
<Tag.Root size="0">
   <Tag.Button>..</Tag.Button>
</Tag.Root>
```

Just don't want to muddy up components with inline logic like this.

cjbell: Ah got it. So how should we resolve though? Introduce a size 0 button?

kylemcd: Yeah I think that makes the most sense, I'm sure we will run across a need for one at some point too.

kylemcd: We can add that as a TODO item in Linear and merge this for now though

kylemcd (approved):

kylemcd:
Looks good, don't forget to add a changeset 👍

PR #161, opened 2024-07-02
feat: vanilla extract css concept

kylemcd:
### Description
- Adds `@telegraph/style-engine` package
  - Light wrapper around `@vanilla-extract/sprinkles` (will use `@vanilla-extract/recipes` for variant workflow when integrating this into other components).
  - `useStyleProps` hook which takes in `props` and the function generated by the style engine `stylePropsFn` to filter `styleProps` from `componentProps` + generate the `styleClassName` too apply to the component. `styleProps` and `componentProps` are fully type safe and reflect the props after filtering (this was a challenge).
  - Exports `RESPONSIVE_STYLE_PROPS` const for use in `defineStyleProps` so we use consistent breakpoints. We could technically wrap `defineStyleProps` and set this to the default, but would seem like a weird override, especially considering there's components we won't use breakpoints in at all. 
- Integrates `@telegraph/style-engine` into `@telegraph/layout` 
   - 1 small breaking change, in vanilla extract responsive style props look like `p={[2,4,6]}` whereas before these responsive styles were handled like `p={{ sm: 2, md: 6 }}`. I can't find a place where we're using this, but noting as it is a change to the component API. 
   - Deletes code from the old way of styling.
- Adds `vanilla-extract` vite plugin to `@telegraph/vite-config` 

### Tasks
[KNO-6322](https://linear.app/knock/issue/KNO-6322/[telegraph]-vanilla-extract-integration)


kylemcd:
* **#161** <a href="https://app.graphite.dev/github/pr/knocklabs/telegraph/161?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

packages/style-engine/src/vite-config.ts
```diff
@@ -0,0 +1,9 @@
+import { vanillaExtractPlugin } from "@vanilla-extract/vite-plugin";
+
+export const styleEngineViteConfig = {
+  plugins: [
+    vanillaExtractPlugin({
+      identifiers: ({ hash }) => `tgph_${hash}`,
```
cjbell, line 6: We have this and `tgph-${hash}` in [style-engine-config.ts](https://github.com/knocklabs/telegraph/pull/161/files#diff-5ee0bad7e3cbdea091c27630187ea010fc5d17bba93a9d99f64a5585338bf2c9)

Should these be aligned?

packages/vite-config/src/vite-config.ts
```diff
@@ -48,6 +48,7 @@ export default {
       external: [...allDependencies],
       output: {
         assetFileNames: (assetInfo: PreRenderedAsset) => {
+          console.log(assetInfo.name);
```
cjbell, line 51: 👀

packages/layout/src/Box/Box.tsx
```diff
@@ -1,64 +1,35 @@
...
   HTMLElement
 > &
-  BoxPropsTokens;
+  StyleProps;
 
 const Box = <T extends TgphElement>({
   as,
   className,
   tgphRef,
+  borderColor = "gray-4",
```
cjbell, line 20: Is this right? Should have this as the default for all boxes?

cjbell (commented):
Few comments

packages/style-engine/src/vite-config.ts
kylemcd: Oh I just need to remove this from `vite-config.ts`

packages/layout/src/Box/Box.tsx
kylemcd: Yeah I think this is a good default so that you don't have to apply the color yourself, instead we just use the color we use as border colors in 99% of places so you can just define the widtha.

cjbell: ohhh got it. yeah agree.

cjbell (approved):
💯 great stuff

kylemcd:
### Merge activity

* **Jul 8, 9:26 AM CDT**: @kylemcd started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/knocklabs/telegraph/161).

PR #208, opened 2024-08-01
feat: kbd component

kylemcd:
### Description
Co-authored with @meryldakin 

- Adds `Kbd` component to display keyboard keys with a pressed state
- Adds `KbdProvider` so that a user can wrap parts of their app in order to limit the amount of event listeners, i.e. no `KbdProvider` means and event listener for each use of `Kbd` vs using `KbdProvider` means 1 event listener for all `Kbd` components. 
- Fixes issue with `useAppearance` props

### Tasks
[KNO-6208](https://linear.app/knock/issue/KNO-6208/[telegraph]-kbd-component-pressed-state)

### DIFF
--- a/packages/kbd/src/Kbd/Kbd.hooks.tsx
+++ b/packages/kbd/src/Kbd/Kbd.hooks.tsx
@@ -1,0 +1,120 @@
+import React from "react";
+
+type PressedContextState = {
+  currentlyPressedKeys: Array<string> | null;
+};
+
+const PressedContext = React.createContext<PressedContextState>({
+  currentlyPressedKeys: null,
+});
+
+type UsePressedParams = {
+  key: KeyboardEvent["key"];
+};
+
+const usePressed = ({ key }: UsePressedParams) => {
+  const context = React.useContext(PressedContext);
+  const [pressed, setPressed] = React.useState(false);
+
+  const handleKeyDown = React.useCallback(
+    (event: KeyboardEvent) => {
+      if (event.key === key) {
+        setPressed(true);
+      }
+    },
+    [key],
+  );
+
+  const handleKeyUp = React.useCallback(
+    (event: KeyboardEvent) => {
+      if (event.key === key) {
+        setPressed(false);
+      }
+    },
+    [key],
+  );
+
+  React.useEffect(() => {
+    // If currentPressedKey is equal to null then this hook
+    // is not contained within a KbdProvider. If it is any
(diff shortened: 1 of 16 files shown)


packages/kbd/src/Kbd/Kbd.hooks.tsx
```diff
@@ -0,0 +1,120 @@
...
+  const handleKeyUp = React.useCallback(
+    (event: KeyboardEvent) => {
+      if (event.key === key) {
+        setPressed(false);
+      }
+    },
+    [key],
+  );
+
+  React.useEffect(() => {
```
meryldakin, line 37: just curious, why do you use `React.useEffect` (and useState etc) instead of importing `useEffect` and using it without the preface?

packages/kbd/src/Kbd/Kbd.hooks.tsx
```diff
@@ -0,0 +1,120 @@
...
+type KbdProviderProps = {
+  children: React.ReactNode;
+};
+
+const KbdProvider = ({ children }: KbdProviderProps) => {
+  const [currentlyPressedKeys, setCurrentlyPressedKeys] = React.useState<
+    PressedContextState["currentlyPressedKeys"]
+  >([]);
+
+  // Brodcast the the keys currently being pressed, passing
```
meryldakin, line 72:
```suggestion
  // Broadcast the the keys currently being pressed, passing
```

packages/kbd/src/Kbd/Kbd.hooks.tsx
```diff
@@ -0,0 +1,120 @@
...
+  }, [context?.currentlyPressedKeys, key, handleKeyDown, handleKeyUp]);
+
+  return { pressed };
+};
+
+type KbdProviderProps = {
+  children: React.ReactNode;
+};
+
+const KbdProvider = ({ children }: KbdProviderProps) => {
```
meryldakin, line 67: cool 😎

meryldakin (approved):

packages/kbd/src/Kbd/Kbd.hooks.tsx
kylemcd: For a lot of these I end up importing a lot of types from react so it's just cleaner to import top level `React` here. Also a personal preference as I find most instances of destructing leading to hard to read/less verbose code.

connorlindsey (approved):
✨

PR #212, opened 2024-08-05
chore: change button and tag rounded radius

cjbell:
> Size 0, 1, and 2 Buttons and Tags should have a border radius of rounded-2 instead of rounded-3


kylemcd (approved):
Looks good, just need to add a changeset 👍

PR #215, opened 2024-08-06
fix: update radio card color styles and spacing

connorlindsey:
Updates the text/icon colors for the unselected, hover, and active states. Also updates the spacing between cards. 

Screenshot 2024-08-06 at 10 47 00 AM

Looks like the built css is correct too:

Screenshot 2024-08-06 at 10 50 02 AM

### DIFF
--- a/packages/radio/src/RadioCards/RadioCards.tsx
+++ b/packages/radio/src/RadioCards/RadioCards.tsx
@@ -27,7 +27,7 @@ const Root = React.forwardRef<RootRef, RootProps>(
           {...props}
           ref={forwardedRef}
         >
-          <Stack gap="2">{children}</Stack>
+          <Stack gap="1">{children}</Stack>
         </RadioGroup.Root>
       </RadioButtonContext.Provider>
     );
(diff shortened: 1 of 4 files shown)


packages/radio/src/RadioCards/RadioCards.tsx
```diff
@@ -74,15 +74,25 @@ const ItemDescription = <T extends TgphElement>({
   size = "0",
   ...props
 }: ItemDescriptionProps<T>) => {
-  return <Button.Text as={"span"} size={size} {...props} />;
+  return (
+    <Button.Text
+      className="tgph-radio-card-description"
```
kylemcd, line 79: Let's make sure we're passing `className` from props into all of these as well 

`className={clsx("tgph-radio-card-description", className)}`

packages/radio/src/default.css
```diff
@@ -1,4 +1,18 @@
...
+
+/* Style hovered or active radio card elements */
+.tgph-radio-card-root:hover .tgph-radio-card-description,
+.tgph-radio-card-root[data-tgph-button-active="true"]
+  .tgph-radio-card-description,
+.tgph-radio-card-root:hover .tgph-radio-card-icon,
+.tgph-radio-card-root[data-tgph-button-active="true"] .tgph-radio-card-icon {
+  color: var(--tgph-gray-12) !important;
+}
+
```
kylemcd, line 17: Can we move this into RadioCards.styles.css and then `@import` that into default css?

kylemcd (approved):

PR #278, opened 2024-10-03
feat: update RadioCards to support direction

mattmikolay:
This PR updates `RadioCards` so that:
- The `RootProps` type reflects the fact that `Stack` props can be passed in. (This allows us to set a `direction` prop on the `RadioCards` component.)
- When the `options` prop is used, the layout of rendered card items will depend upon the `direction` of the root `RadioCards` component.

## Video

https://github.com/user-attachments/assets/8b3bea65-1b57-40b6-8a8c-dc7d469e08b3

### DIFF
--- a/packages/radio/src/RadioCards/RadioCards.tsx
+++ b/packages/radio/src/RadioCards/RadioCards.tsx
@@ -114,21 +114,22 @@ type DefaultProps = React.ComponentPropsWithoutRef<typeof Root> & {
 };
 
 const Default = ({ options, direction = "row", ...props }: DefaultProps) => {
+  const isHorizontal = direction === "row" || direction === "row-reverse";
   return (
     <Root direction={direction} {...props}>
       {options.map((option) => (
         <Item value={option.value}>
           <Stack
-            direction={direction === "row" ? "column" : "row"}
-            align={direction === "row" ? "flex-start" : "center"}
+            direction={isHorizontal ? "column" : "row"}
+            align={isHorizontal ? "flex-start" : "center"}
             p="3"
             w="full"
           >
             {option.icon && (
               <Box
-                mb={direction === "row" ? "2" : "0"}
-                mr={direction === "row" ? "0" : "4"}
-                ml={direction === "row" ? "0" : "1"}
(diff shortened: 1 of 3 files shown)


packages/radio/src/RadioCards/RadioCards.tsx
```diff
@@ -8,7 +8,8 @@ import React from "react";
 
 import { baseStyles } from "./RadioCards.css";
 
-type RootProps = React.ComponentPropsWithoutRef<typeof RadioGroup.Root>;
+type RootProps = React.ComponentPropsWithoutRef<typeof RadioGroup.Root> &
+  Omit<TgphComponentProps<typeof Stack>, "tgphRef">;
```
mattmikolay, line 12: We should omit `Stack`’s `tgphRef` prop here because the `Root` component in this file uses `forwardRef` — correct?

packages/radio/src/RadioCards/RadioCards.tsx
```diff
@@ -112,14 +113,24 @@ type DefaultProps = React.ComponentPropsWithoutRef<typeof Root> & {
...
-const Default = ({ options, ...props }: DefaultProps) => {
+const Default = ({ options, direction = "row", ...props }: DefaultProps) => {
+  const isHorizontal = direction === "row" || direction === "row-reverse";
   return (
-    <Root {...props}>
+    <Root direction={direction} {...props}>
       {options.map((option) => (
-        <Item value={option.value}>
-          <Stack direction="column" align="flex-start" p="3" w="full">
+        <Item key={option.value} value={option.value}>
```
mattmikolay, line 121: I added a missing `key` prop here to fix a warning.

packages/radio/src/RadioCards/RadioCards.tsx
kylemcd: No I think we'd want to keep it because tgphRef eventually ends up down to Stack which is where it's useful. We actually need to remove forwardRef from here entirely.

kylemcd (commented):
Looks good! One more change while you're in here and we're good to go 👍

kylemcd (approved):

PR #283, opened 2024-10-09
feat: various tooltip improvements and fixes

kylemcd:
### Tooltip appearance
- Allows an `appearance` prop to be set on a tooltip with the default set to `dark`. This helps to support cases where we want to override the theme of the tooltip like in the `DateTimePopover`
- Adds an `OverrideAppearance` helper component to `@telegraph/appearance` to support this work. 
- Adds the `appearance` prop to storybook to test this behavior.

### Disabled state handling for button + tooltip
- Previously, we relied on `pointer-events: none` when `disabled` was set to block all `<Button/>` click events. This worked great for that, but not so great when we wanted things like hover events to trigger tooltips. This PR fixes this issue by removing the `pointer-events: auto` style and instead relies on the native button disabled state. For example, when passing `as={Link}` _or_ `as="a"` along with `disabled={true}`, we now override the `as` prop and replace it with `button`. This works splendidly because it still prevents clicks BUT still allows hover events for our tooltips.
- The tooltip now detects if there the children within it are disabled, and if so changes the delay to 0. This helps to better support the pattern throughout control of letting the user know why a button is disabled. 
- Adds disabled button + tooltip examples to storybook to test this behavior. Also, adds a test in `Button` to ensure that a disabled `as="a"` gets transformed into a button.

### Misc
- Lowers the default delay duration for tooltips to 400ms as the previous time set (600ms) felt too slow.

## Tasks
- [KNO-7085](https://linear.app/knock/issue/KNO-7085/[telegraph]-adjust-tooltip-delay-duration-timing) 
- [KNO-6986 ](https://linear.app/knock/issue/KNO-6986/[telegraph]-button-used-as-link-disabled-state-broken)
- [KNO-7008 ](https://linear.app/knock/issue/KNO-7008/[telegraph]-improve-tooltip-disabled-state-interaction)
- [KNO-7116](https://linear.app/knock/issue/KNO-7116/[telegraph]-allow-theme-to-be-passed-to-tooltip)

### DIFF
--- a/.changeset/dry-peas-travel.md
+++ b/.changeset/dry-peas-travel.md
@@ -1,0 +1,7 @@
+---
+"@telegraph/appearance": patch
+"@telegraph/tooltip": patch
+"@telegraph/button": patch
+---
+
+support disabled state in tooltip + tooltip appearance override
--- a/packages/appearance/src/index.ts
+++ b/packages/appearance/src/index.ts
@@ -1,1 +1,6 @@
-export { useAppearance, Appearance, InvertedAppearance } from "./useAppearance";
+export {
+  useAppearance,
+  Appearance,
+  InvertedAppearance,
+  OverrideAppearance,
+} from "./useAppearance";
--- a/packages/appearance/src/useAppearance.tsx
+++ b/packages/appearance/src/useAppearance.tsx
@@ -112,6 +112,22 @@ const Appearance = ({
   return <Component {...derivedAppearanceProps} {...props} />;
 };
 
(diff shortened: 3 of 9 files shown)


kylemcd:
* **#283** <a href="https://app.graphite.dev/github/pr/knocklabs/telegraph/283?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

mattmikolay (approved):

PR #286, opened 2024-10-16
fix: only render children when Tooltip is disabled

connorlindsey:
Removes the extra span from the disabled tooltip. Seems to be causing minor layout changes when the tooltip was enabled/disabled with the following markup:

```
<Stack>
  <Button />
  <Tooltip enabled={isEnabled} label={...}>
    <Button />
  </Tooltip>
</Stack>
```
When disabled, the span was adding 2px around the right button. Cannot track down why that is (there's no border, etc. - the span shouldn't be affecting anything), but I think not adding the extra span is cleaner regardless.
CleanShot 2024-10-16 at 05 25 49

### DIFF
--- a/.changeset/rude-lions-fly.md
+++ b/.changeset/rude-lions-fly.md
@@ -1,0 +1,5 @@
+---
+"@telegraph/tooltip": patch
+---
+
+fix: only render children when Tooltip is disabled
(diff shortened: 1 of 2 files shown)


kylemcd (approved):
I think this change is fine but there's definitely something else at play in your example picture. Have done this same exact pattern a few times now and haven't ran into it. Would make sure you have a flex box with items aligned center.

PR #299, opened 2024-10-29
feat: select component

kylemcd:
### Description
- Adds `<Select/>` component. A simpler version of `<Combobox/>` that more closely matches the browser `<select/>` element.
- Support the `multiple` prop like the browser `<select/>`
- Fixes [KNO-6849](https://linear.app/knock/issue/KNO-6849/[telegraph]-allow-combobox-labels-to-be-react-nodes) so that we can pass react nodes to the label.
- Fixes [KNO-6934](https://linear.app/knock/issue/KNO-6934/[telegraph]-allow-disabled-state-without-type-error-on-combobox) so that we can pass the `disabled` prop to the select component and combobox component.

### Tasks
[KNO-7249](https://linear.app/knock/issue/KNO-7249/[telegraph]-select-component)
[KNO-6934](https://linear.app/knock/issue/KNO-6934/[telegraph]-allow-disabled-state-without-type-error-on-combobox)
[KNO-6849](https://linear.app/knock/issue/KNO-6849/[telegraph]-allow-combobox-labels-to-be-react-nodes)

### DIFF
--- a/packages/select/src/Select/Select.tsx
+++ b/packages/select/src/Select/Select.tsx
@@ -1,0 +1,81 @@
+import { Combobox } from "@telegraph/combobox";
+import { TgphComponentProps } from "@telegraph/helpers";
+import React from "react";
+
+type Option = TgphComponentProps<typeof Combobox.Option>;
+
+type RootProps = Omit<
+  TgphComponentProps<typeof Combobox.Root>,
+  "value" | "onValueChange" | "layout"
+> & {
+  value?: Option["value"];
+  onValueChange?: (value: Option["value"]) => void;
+  size?: TgphComponentProps<typeof Combobox.Trigger>["size"];
+  triggerProps?: TgphComponentProps<typeof Combobox.Trigger>;
+  contentProps?: TgphComponentProps<typeof Combobox.Content>;
+  optionsProps?: TgphComponentProps<typeof Combobox.Options>;
+};
+
+const Root = ({
+  size = "1",
+  value,
+  onValueChange,
+  triggerProps,
+  contentProps,
+  optionsProps,
+  children,
+  ...props
+}: RootProps) => {
+  const [options, setOptions] = React.useState<Array<Option>>([]);
+
+  // Get all of the options passed into Select so that we can display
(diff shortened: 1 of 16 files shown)


kylemcd:
* **#299** <a href="https://app.graphite.dev/github/pr/knocklabs/telegraph/299?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

packages/select/src/Select/Select.tsx
```diff
@@ -0,0 +1,121 @@
...
+  multiple?: boolean;
+} & (
+    | {
+        multiple?: true;
+        value?: MultipleValue;
+        onValueChange?: (value: MultipleValue) => void;
+        layout?: TgphComponentProps<typeof Combobox.Root>["layout"];
+      }
+    | {
+        multiple?: false | undefined;
```
mattmikolay, line 27: I think the `| undefined` should be unnecessary since this is already an optional property: `multiple?`.

packages/select/src/Select/Select.tsx
```diff
@@ -0,0 +1,121 @@
...
+  multiple = false,
+  value,
+  onValueChange,
+  triggerProps,
+  contentProps,
+  optionsProps,
+  children,
+  ...props
+}: RootProps) => {
+  const [options, setOptions] = React.useState<Array<Option>>([]);
```
mattmikolay, line 45: Is there an issue that occurs if you assign `options` without using `useState` and `useEffect`? I’m wondering if you can just move all this logic into an `options = useMemo(() => ...)`.

mattmikolay (approved):
Looks great! 🚢  One nit and a question!

packages/select/src/Select/Select.tsx
kylemcd: Ah good catch, didn't delete this part when debugging.

packages/select/src/Select/Select.tsx
kylemcd: Yeah good call 👍

connorlindsey (approved):
🙌

PR #313, opened 2024-11-06
fix: improve modal stacking with portal

kylemcd:
### Description
- Changes the modal structure to always house modals inside of a `Portal`.
- Updates how `FocusScope` was added to the component allowing for better control of trapping and allowing for props to be passed to it. 

### Tasks
[KNO-7292](https://linear.app/knock/issue/KNO-7292/[telegraph]-overlapping-modal-w-code-editor-is-broken)

### DIFF
--- a/.changeset/six-turtles-scream.md
+++ b/.changeset/six-turtles-scream.md
@@ -1,0 +1,5 @@
+---
+"@telegraph/modal": patch
+---
+
+fix some modal stacking issues by making modals exist in portals
--- a/packages/modal/src/Modal/Modal.stories.tsx
+++ b/packages/modal/src/Modal/Modal.stories.tsx
@@ -34,7 +34,7 @@ type Story = StoryObj<typeof TelegraphModal.Root>;
 export const Modal: Story = {
   render: () => {
     // eslint-disable-next-line react-hooks/rules-of-hooks
-    const [open, setOpen] = React.useState(true);
+    const [open, setOpen] = React.useState(false);
     const Modal = TelegraphModal;
     return (
       <>
@@ -66,7 +66,7 @@ export const Modal: Story = {
 export const ScrollingModal: Story = {
   render: () => {
     // eslint-disable-next-line react-hooks/rules-of-hooks
-    const [open, setOpen] = React.useState(true);
+    const [open, setOpen] = React.useState(false);
     const Modal = TelegraphModal;
     return (
       <>
@@ -204,12 +204,19 @@ const NestedModal = ({
   onOpenChange,
   onClick,
 }: NestedModalProps) => {
+  const buttonRef = React.useRef<HTMLButtonElement>(null);
   return (
     <>
       <TelegraphModal.Root
         a11yTitle={`Nested Modal ${depth}`}
         open={open}
         onOpenChange={onOpenChange}
+        onMountAutoFocus={(event) => {
+          // Demonstrate focusing on a specific element
+          // when the modal is opened
+          event.preventDefault();
+          buttonRef.current?.focus();
+        }}
       >
         <TelegraphModal.Content>
           <TelegraphModal.Header>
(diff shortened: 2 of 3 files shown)


kylemcd:
* **#313** <a href="https://app.graphite.dev/github/pr/knocklabs/telegraph/313?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

packages/modal/src/Modal/Modal.stories.tsx
```diff
@@ -34,7 +34,7 @@ type Story = StoryObj<typeof TelegraphModal.Root>;
 export const Modal: Story = {
   render: () => {
     // eslint-disable-next-line react-hooks/rules-of-hooks
-    const [open, setOpen] = React.useState(true);
+    const [open, setOpen] = React.useState(false);
```
kylemcd, line 37: Changed these to false by default so they don't display over all of the storybook content now that they're in portals.

packages/modal/src/Modal/Modal.tsx
```diff
@@ -45,18 +47,30 @@ const Root = ({
...
   open,
   onOpenChange,
   a11yTitle,
   a11yDescription,
   children,
   layer: layerProp,
+  // Focus scope props
+  trapped,
+  onMountAutoFocus,
+  onUnmountAutoFocus,
```
kylemcd, line 65: Allowed for FocusScope props to be passed so we can have better control of the focused element inside the modal.

packages/modal/src/Modal/Modal.tsx
```diff
@@ -45,18 +47,30 @@ const Root = ({
...
   ...props
-}: Required<RootProps, "open" | "onOpenChange">) => {
-  const id = React.useId();
+}: RootComponentProps) => {
+  // We use the a11yTitle as the id for the modal as it is unique
+  // and can be used to identify the modal in the stacking context.
+  // Without the need to generate a random id and manage between
+  // different modal rendering patterns.
+  const id = a11yTitle;
```
kylemcd, line 72: I originally used `React.useId()` here. This worked fine when modals were "nested" side by side, i.e.

```
<Modal></Modal>
<Modal></Modal>
<Modal></Modal>
```

But when actually nesting modals inside of each other like this:

```
<Modal>
   <Modal>
      <Modal>
      </Modal>
    </Modal>
</Modal
```

The id generated from `useId()` would update on re-render. This meant that there was no consistent key that we could use to manage the layers within the stacking context. 

Fortunately, the `a11yTitle` prop is a required prop that is meant to be unique. This means that we can have a consistent key that works outside of re-renders. So making this change fixed some of the modal stacking issues I was seeing after utilizing a portal.

packages/modal/src/Modal/Modal.tsx
```diff
@@ -111,51 +126,68 @@ const RootComponent = ({
...
-                  w={props.w ?? "full"}
-                  bg="surface-1"
-                  border="px"
-                  rounded="4"
-                  shadow="3"
-                  key={`content-${id}`}
-                  {...props}
+            // We add the className "tgph" here so that styles within
+            // the portal get scoped properly to telegraph
+            <Portal.Root className="tgph">
```
kylemcd, line 131: Added `Portal.Root` here. This makes it so every modal will exist in a portal and is much easier to manage for stacking as each modal is now at the same hierarchy in the DOM. I don't see any use case that we wouldn't want to use a portal here so I made it the default.

packages/modal/src/Modal/Modal.tsx
```diff
@@ -111,51 +126,68 @@ const RootComponent = ({
...
+              <Overlay layer={layer}>
+                <FocusScope
+                  trapped={typeof trapped === "boolean" ? trapped : isTopLayer}
+                  onMountAutoFocus={onMountAutoFocus}
+                  onUnmountAutoFocus={onUnmountAutoFocus}
+                  asChild
```
kylemcd, line 137: Moved FocusScope into an area I thought made more sense. Also added the ability to pass FocusScope props.

mattmikolay (approved):
🚢

.changeset/six-turtles-scream.md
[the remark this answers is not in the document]
kylemcd: hahaha that's incredible

PR #351, opened 2025-01-21
feat: rework style engine and apply to layout / typography

kylemcd:
### Summary of Changes

1. **New Style Engine**
   - **CSS Variables**: The new style engine uses scoped CSS variables (`--tgph-*`) to dramatically cut down CSS bundle size while making styles reusable and dynamic.
   - **Core Pieces**:
     - **`useStyleEngine` Hook**: A React hook that maps component props to CSS variables from a `cssVars` configuration and applies them as inline styles.
     - **`getStyleProp` Utility**: A utility that does the same mapping but works outside React, giving you flexibility to handle styles wherever you need them.

2. **Refactor of `@telegraph/layout` and `@telegraph/typography`**
   - These packages were reworked to use the new style engine since they were the main offenders in bloating CSS bundles.
   - Tailwind utilities have been removed, and styles are now simpler and easier to maintain.
   - Other components can be migrated to this new setup incrementally, no need to do it all at once.

3. **Testing Updates**
   - Tests now validate CSS variables directly instead of asserting Tailwind classes. This ensures tests are more focused on the actual implementation details and helps avoid regressions with the new architecture.

---

#### Example

Here’s how you’d define values to pass to the `useStyleEngine` hook:

```ts
import { useStyleEngine } from "@telegraph/style-engine";

const cssVars = {
  color: {
    cssVar: "--color",
    value: "var(--tgph-VARIABLE)",
  },
  padding: {
    cssVar: "--padding",
    value: "var(--tgph-spacing-VARIABLE)",
  },
};

const { styleProp, otherProps } = useStyleEngine({
  props: { color: "red-11", padding: "4" },
  cssVars,
});
```

### Tasks
KNO-7681


kylemcd:
* **#351** <a href="https://app.graphite.dev/github/pr/knocklabs/telegraph/351?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/knocklabs/telegraph/351?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/style-engine/src/helpers/getStyleProp.ts
```diff
@@ -0,0 +1,195 @@
...
+  props: Props & { style?: Record<string, string> };
+  cssVars: CssVars;
+};
+
+export const getStyleProp = <
+  CssVars extends CssVarsPropObject<CssVars>,
+  Props extends Record<string, unknown>,
+>(
+  params: GetStylePropParams<CssVars, Props>,
+): {
```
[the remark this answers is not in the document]
kylemcd: Yes def!

packages/style-engine/src/hooks/useStyleEngine.ts
```diff
@@ -0,0 +1,12 @@
...
+import { type CssVarsPropObject, getStyleProp } from "../helpers/getStyleProp";
+
+export const useStyleEngine = <
+  CssVars extends CssVarsPropObject<CssVars>,
+  Props extends Record<string, unknown>,
+>(
+  params: Parameters<typeof getStyleProp<CssVars, Props>>[0],
+) => {
+  return React.useMemo(() => getStyleProp(params), [params]);
+};
```
cjbell, line 12: Have we considered the performance characteristics of this yet? Given that we're going to be doing a bunch of this at runtime now and not buildtime, I think we might need to dig in here more before we ship this. 

The reason I'm calling this out especially here is that because we're going to be defining the object `params` in the component that means that the shallow equality done by the useMemo will always return false, meaning that this hook will **always** recalculate.

cjbell (commented):
one very important performance question

packages/style-engine/src/hooks/useStyleEngine.ts
kylemcd: It's a pretty light function to run but maybe there's a better way to recognize a relevant prop change. Will dig into this a bit deeper.

kylemcd: @cjbell I think this is more than fine. I put 2k+ buttons on a storybook page, where each button includes 4 components using the hook and it changed the color on all buttons instantly. The performance on the slowest run of one of the functions is 0.1ms (0.0001 seconds).

Let me know what you think but I don't have any perf concerns here.

PR #372, opened 2025-01-28
fix: wrap combobox content in portal

connorlindsey:
Wraps combobox content in radix portal like we do with the modal to fix overlapping issues when within other components like popovers.

CleanShot 2025-01-28 at 11 38 22


connorlindsey:
Closing in favor of the work Kyle is doing on the combobox

PR #393, opened 2025-02-07
fix(KNO-7736): support setting maxHeight on Combobox.Options

mattmikolay:
This PR updates `Combobox.Options` so that the default stack max height of `calc(var(--tgph-combobox-content-available-height) - var(--tgph-spacing-12))` is only applied when a custom `maxHeight` prop is not passed in.

### DIFF
--- a/packages/combobox/src/Combobox/Combobox.tsx
+++ b/packages/combobox/src/Combobox/Combobox.tsx
@@ -721,7 +721,7 @@ const Options = <T extends TgphElement>({ ...props }: OptionsProps<T>) => {
       style={{
         overflowY: "auto",
         // maxHeight defaults to available height - padding from edge of screen
-        maxHeight: !maxHeight
+        "--max-height": !maxHeight
           ? "calc(var(--tgph-combobox-content-available-height) - var(--tgph-spacing-12))"
           : undefined,
       }}
--- a/packages/combobox/src/Combobox/Combobox.tsx
+++ b/packages/combobox/src/Combobox/Combobox.tsx
@@ -711,8 +711,6 @@ type OptionsProps<T extends TgphElement> = TgphComponentProps<typeof Stack<T>>;
 const Options = <T extends TgphElement>({ ...props }: OptionsProps<T>) => {
   const context = React.useContext(ComboboxContext);
 
-  const { maxHeight } = props;
-
   return (
     <Stack
       id={context.contentId}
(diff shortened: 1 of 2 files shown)


mattmikolay:
* **#393** <a href="https://app.graphite.dev/github/pr/knocklabs/telegraph/393?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/knocklabs/telegraph/393?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/combobox/src/Combobox/Combobox.tsx
```diff
@@ -707,16 +707,19 @@ type OptionsProps<T extends TgphElement> = TgphComponentProps<typeof Stack<T>>;
...
       gap="1"
       style={{
         overflowY: "auto",
-        // Available Height - Padding from edge of screen
-        maxHeight:
-          "calc(var(--tgph-combobox-content-available-height) - var(--tgph-spacing-12))",
+        // maxHeight defaults to available height - padding from edge of screen
+        maxHeight: !maxHeight
+          ? "calc(var(--tgph-combobox-content-available-height) - var(--tgph-spacing-12))"
+          : undefined,
```
kylemcd, line 722: Can we instead change this to utilize the `maxHeight` prop that the `<Stack/>` component already supports?. i.e. If there is no `maxHeight / maxH` prop passed here, then set the css var via style tag like this:

```
style={{ "--max-height": "calc(var(--tgph-combobox-content-available-height) - var(--tgph-spacing-12))" }}
```

If it is passed then pass the `maxHeight / maxH` prop directly to the `<Stack/>` like usual while not passing `maxHeight` css directive to the `style` prop.

packages/combobox/src/Combobox/Combobox.tsx
```diff
@@ -707,16 +707,19 @@ type OptionsProps<T extends TgphElement> = TgphComponentProps<typeof Stack<T>>;
...
       id={context.contentId}
       direction="column"
       gap="1"
       style={{
         overflowY: "auto",
-        // Available Height - Padding from edge of screen
-        maxHeight:
-          "calc(var(--tgph-combobox-content-available-height) - var(--tgph-spacing-12))",
+        // maxHeight defaults to available height - padding from edge of screen
+        "--max-height": !maxHeight
```
kylemcd, line 720: Probably makes sense to access maxHeight as `props.maxHeight` and then just pass props to Stack as usual. Since right now in this setup the `maxHeight` prop isn't getting passed onto the Stack.

kylemcd (commented):
One small change to make sure `maxHeight` is passed through correctly to Stack.

packages/combobox/src/Combobox/Combobox.tsx
mattmikolay: I actually destructure on line 710:

```tsx
const { maxHeight } = props;
```

But I agree this is not so clear. I’m updating this to use `props.maxHeight`. 👍🏻

kylemcd (approved):

PR #396, opened 2025-02-10
fix(KNO-7737): Long items in combobox and menu options do not overflow

MikeCarbone:
#  fix(KNO-7737): Long items in combobox and menu options do not overflow

[Linear Issue](https://linear.app/knock/issue/KNO-7737/[telegraph]-value-of-single-option-combobox-does-not-ellipsize)

This PR fixes the combobox value and the option in the menu from overflowing its container:

Screenshot 2025-02-10 at 10 22 42 AM

## Handling menu items too

This PR also applies the fix to menu item options. It was discovered that long items of text will overflow its container in the menu.

### Before:
Screenshot 2025-02-10 at 10 14 48 AM

### After:
Screenshot 2025-02-07 at 4 54 42 PM
Screenshot 2025-02-10 at 10 16 07 AM

##


packages/combobox/src/Combobox/Combobox.tsx
```diff
@@ -502,7 +508,7 @@ const Trigger = ({ size = "2", ...props }: TriggerProps) => {
         aria-haspopup="listbox"
         // Custom attributes
         data-tgph-combobox-trigger
-        data-tgph-comobox-trigger-open={context.open}
+        data-tgph-combobox-trigger-open={context.open}
```
mattmikolay, line 511: Nice catch! 😎

packages/combobox/src/Combobox/Combobox.tsx
```diff
@@ -379,7 +379,13 @@ const TriggerValue = () => {
       context.legacyBehavior && (context?.value as DefinedOption)?.label;
 
     return (
-      <TelegraphButton.Text color={!context.value ? "gray" : "default"}>
+      <TelegraphButton.Text
+        color={!context.value ? "gray" : "default"}
+        style={{
+          textOverflow: "ellipsis",
+          overflow: "hidden",
+        }}
```
mattmikolay, line 387: So Telegraph actually has a pretty flexible styling system already in place. Please double check me, but I think you can already set overflow just by passing a prop to `<TelegraphButton.Text>`:

```tsx
<TelegraphButton.Text overflow="hidden" />
```

It would be nice to make the same thing possible for `textOverflow`; I don’t believe we already have that ability. You'll need to update the style constants and `baseCssVars` associated with Telegraph’s typography:

https://github.com/knocklabs/telegraph/blob/c33850a7131824bfaf8fdb41808f853c9399cdad/packages/typography/src/constants.ts#L4-L12

mattmikolay (commented):
Nice work! I have one piece of feedback based on how styling works in Telegraph.

packages/menu/src/MenuItem/MenuItem.tsx
```diff
@@ -44,7 +44,18 @@ const MenuItem = <T extends TgphElement>({
...
+          // Styles to prevent long items from overflowing the menu
+          // max-width makes sure the text maintains the correct left alignment
+          style={{
+            maxWidth: "90%",
+            textOverflow: "ellipsis",
+            overflow: "hidden",
+          }}
+        >
+          {props.children}
+        </Button.Text>
```
kylemcd, line 58: This can be done without setting an explicit `maxWidth` which I think would be a code smell?

```tsx
      <Stack gap={gap} align="center" w="full">
        <MenuItemLeading
          icon={icon}
          selected={selected}
          leadingIcon={leadingIcon}
          leadingComponent={leadingComponent}
        />
        <Button.Text
          weight="regular"
          overflow="hidden"
          style={{
            textOverflow: "ellipsis",
          }}
        >
          {props.children}
        </Button.Text>
      </Stack>
```

MikeCarbone: Screenshot 2025-02-10 at 11 52 12 AM

Agreed not great but without the max-width the left alignment breaks on the option. The element starts pushing leftwards as it expands out. Not sure if there's a better way to accomplish this?

packages/combobox/src/Combobox/Combobox.tsx
MikeCarbone: Good call out. I'm wrapping my head around these styling patterns, might have some Qs here

packages/menu/src/MenuItem/MenuItem.tsx
kylemcd: Ah I see. If you add `w="full"` to the `MenuItem` `Button.Text` and then add `display="block"` to the `Button.Icon` inside of `MenuItemLeading` that should sort things out.

MikeCarbone: Oh yeah that'll do it! Good stuff

kylemcd (approved):
Looks great! Amazing job!

PR #397, opened 2025-02-10
Search using children coerced to text

mattmikolay:


mattmikolay:
* **#397** <a href="https://app.graphite.dev/github/pr/knocklabs/telegraph/397?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/knocklabs/telegraph/397?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/combobox/src/Combobox/Combobox.helpers.ts
```diff
@@ -107,3 +107,24 @@ export const doesOptionMatchSearchQuery = ({
...
   );
 };
+
+export const coerceChildrenToText = (
+  children: React.ReactNode,
+): string | undefined => {
+  const childrenArray = React.Children.toArray(children);
+
+  return childrenArray
+    .map((child) => {
```
mattmikolay, line 117: Instead of returning the first string encountered, I chose to take this approach because it better supports cases such as the following:

```tsx
(
  <Combobox.Option value={value}>
    <span>Title</span>
    <span>Subtitle</span>
  </Combobox.Option>
)
```

mattmikolay: I also considered adding a ref to the DOM element rendered by `<Option>`, to use the element’s `innerText`, but that creates a weird scenario whereby the element’s `innerText` determines its visibility (via `isVisible`), yet its `innerText` also depends on its visibility. It worked when I tested things, but that kind of cyclical logic feels dangerous to me.

packages/combobox/src/Combobox/Combobox.helpers.ts
```diff
@@ -107,3 +107,24 @@ export const doesOptionMatchSearchQuery = ({
...
+      if (typeof child === "string") {
+        return child;
+      }
```
MikeCarbone, line 120: would this logic still work if a child is something like `<p>NJ <span>(New Jersey)</span></p>`?

packages/combobox/src/Combobox/Combobox.tsx
```diff
@@ -749,10 +750,15 @@ const Option = <T extends TgphElement>({
...
+  const textContent = React.useMemo(
+    () => coerceChildrenToText(children),
+    [children],
+  );
+
   const isVisible =
     !context.searchQuery ||
     doesOptionMatchSearchQuery({
-      label: label?.toString() || children?.toString(),
+      label: label?.toString() || textContent,
```
MikeCarbone, line 761: if this is an OR condition based on the label existing, should we do this check before running the useMemo above? Otherwise the `coerceChildrenToText` function would run each time, even if we don't have to

packages/combobox/src/Combobox/Combobox.helpers.ts
```diff
@@ -107,3 +107,24 @@ export const doesOptionMatchSearchQuery = ({
     label?.toLowerCase().includes(searchQuery.toLowerCase())
   );
 };
+
+export const coerceChildrenToText = (
```
kylemcd, line 111: I think it might make more sense here to do logic like:

`if string -> push to array`, then compare text from search to each string in the array?

You'll also want to add a `foundChildren` var to the function that you initialize as an empty array that you would push those strings to vs returning the single depth of the `childrenArray`.

packages/combobox/src/Combobox/Combobox.tsx
kylemcd, line 761: I think I'd like to see `textContent` derived within `doesOptionMatchSearchQuery` instead of within this component.

packages/combobox/src/Combobox/Combobox.helpers.ts
mattmikolay: Yes — I just added some UTs to test this. 🙂

packages/combobox/src/Combobox/Combobox.tsx
```diff
@@ -752,7 +752,7 @@ const Option = <T extends TgphElement>({
   const isVisible =
     !context.searchQuery ||
     doesOptionMatchSearchQuery({
-      label: label?.toString() || children?.toString(),
+      label: label || children,
```
MikeCarbone, line 755: should we change this parameter name? Feels like using the word `label` here and within `doesOptionMatchSearchQuery` might be outdated / nondescript now that we pass children in too

PR #399, opened 2025-02-10
chore(KNO-7827): adds instructions to readme for developing locally

MikeCarbone:


kylemcd (approved):
Looks great! You can remove the changset since we're not actually bumping any versions with this!

meryldakin (approved):

PR #426, opened 2025-03-04
fix(KNO-7992): Fix margin and marginX in Box constants

mattmikolay:
I noticed some weirdness when adding an `mb` prop to the `<Heading>` component. This fixes it.

### DIFF
--- a/packages/layout/src/Box/Box.constants.ts
+++ b/packages/layout/src/Box/Box.constants.ts
@@ -160,10 +160,12 @@ const baseCssVars: Record<keyof BaseStyleProps, CssVarProp> = {
   margin: {
     cssVar: "--margin",
     value: "var(--tgph-spacing-VARIABLE)",
+    direction: "all",
   },
   marginX: {
     cssVar: "--margin",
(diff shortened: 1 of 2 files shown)


packages/layout/src/Box/Box.constants.ts
```diff
@@ -160,10 +160,12 @@ const baseCssVars: Record<keyof BaseStyleProps, CssVarProp> = {
   margin: {
     cssVar: "--margin",
     value: "var(--tgph-spacing-VARIABLE)",
+    direction: "all",
   },
   marginX: {
     cssVar: "--margin",
-    value: "0 var(--tgph-spacing-VARIABLE)",
+    value: "var(--tgph-spacing-VARIABLE)",
+    direction: "x",
```
mattmikolay, line 168: Here are `padding` and `paddingX` for reference:

https://github.com/knocklabs/telegraph/blob/d1f4a5127a8857d98ef7cfea5e50530f521158c6/packages/layout/src/Box/Box.constants.ts#L125-L134

PR #456, opened 2025-05-23
feat: decrease icon package downstream bundle size by using lucide dynamic

kylemcd:
### Description
The `@telegraph/icon` package exported ALL of the icons from `lucide-react` under the `Lucide` object. In doing the export in this way, a downstream bundler would be unable to successfully tree shake the package making it so ALL of the icons were in the main bundle. This PR utilizes [Lucide's Dynamic Icon Component](https://lucide.dev/guide/packages/lucide-react#dynamic-icon-component) so that during telegraph's bundling process each icon is put into its own file making it so a downstream bundler does not have to treeshake telegraph. This PR also maintains backwards compatibility with the previous `Lucide.Bell` pattern so that no code updates will need to be made by the downstream user.

In this PR we also:
- Update the tests to work with `DynamicIcon`, which required a mocking process because `DynamicIcon` is an ES Module and `vitest` expects to be able to work with CJS.
- Update our `axe` utility to throw the actual error from the violation check so that we can see it in the test output.
- Remove the `throw` for icon component issues in favor of a `console.error` to avoid async issues with the `DynamicIcon` component. This also required adjusting the tests within `@telegraph/icon` and `@telegraph/button`.

As an example, here is our bundle before and after for the dashboard. Notice how before how `@telegraph/icon` was within the main bundle and taking up a large chunk of that. In the after, the `@telegraph/icon` is MUCH smaller as each icon exists in it's own chunk and not in the main bundle. 

Before After Bundle Size

### DIFF
--- a/.changeset/chilled-boxes-check.md
+++ b/.changeset/chilled-boxes-check.md
@@ -1,0 +1,5 @@
+---
+"@telegraph/icon": patch
+---
+
+Changes the export strategy for Lucide icons within @telegraph/icon so that they are individually chunked for easier tree shaking by a downstream user.
--- a/packages/button/src/Button/Button.test.tsx
+++ b/packages/button/src/Button/Button.test.tsx
@@ -32,10 +32,11 @@ describe("Button", () => {
     expectToHaveNoViolations(await axe(container));
   });
   it("icon button without alt is inaccessible", async () => {
-    expect(() =>
-      // @ts-expect-error Testing error case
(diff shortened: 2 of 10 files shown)


kylemcd:
* **#456** <a href="https://app.graphite.dev/github/pr/knocklabs/telegraph/456?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/knocklabs/telegraph/456?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #517, opened 2025-07-15
chore: update color, text, leading, and spacing tokens

ksorathia:
Updating our colors to a custom palette to align with Knock branding.
TODO - update yellow palette to ensure accessibility

### DIFF
--- a/packages/tokens/src/tokens/color.ts
+++ b/packages/tokens/src/tokens/color.ts
@@ -1,16 +1,10 @@
 export const color = {
   transparent: "transparent",
-  white: "#ffffff",
+  white: "#FFFFFF",
   black: "#000000",
   surface: {
-    1: {
-      light: "#ffffff",
-      dark: "#18191b",
-    },
-    2: {
-      light: "#f9f9f8",
-      dark: "#111110",
-    },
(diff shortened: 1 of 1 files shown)


packages/tokens/src/tokens/spacing.ts
```diff
@@ -1,9 +1,13 @@
 export const spacing = {
   px: "1px",
   0: "0px",
+  0.5: "0.125rem",
   1: "0.25rem",
+  1.5: "0.375rem",
   2: "0.5rem",
+  2.5: "0.625rem",
   3: "0.75rem",
+  3.5: "0.875rem",
```
kylemcd, line 10: @ksorathia can you update these to use underscores rather than periods `0_5`. Reasoning for this is that css variables don't support periods and this will downstream get added to them there.

.changeset/gentle-cheetahs-glow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@telegraph/tokens": minor
+---
+
+Add half steps to small spaces. Adjust text size and leading.
```
kylemcd, line 5: Then one last thing, you can combine all of the updates to the tokens package in one of these files vs adding multiple changesets for each packge.

PR #526, opened 2025-07-24
docs: standardize component readme files

kylemcd:
### Description
Our robot overlords at cursor helped us write some *top tier* documentation for our telegraph components. Each package is now properly documented. 

Disclosure: I vetted most of this, but there might be some hallucinations here. Fixed the few obvious ones I found.


connorlindsey (approved):
🖊️

PR #571, opened 2025-09-19
fix(KNO-8096): Do not override button "type" prop typing in telegraph button

andy-knock:
In trying to upgrade the control repo, I found a usecase where we need Button's type prop to be fully compatible with HTML button's type prop.

Also, I forgot to document the prop in readme

**Testing:**
- Add 2 more test cases for comprehensive coverage of 4 potential types

**Verify this package doesn't break control**
In telegraph
`yarn build:packages`
In button directory
`yarn pack`

(copy tgz file version of button to control)

In control
Update `package.json` for `@telegraph/button` to be `"@telegraph/button": "file:package.tgz"`
`yarn install`
`yarn tc` 

No type errors
Screenshot 2025-09-19 at 11 11 58 AM

### DIFF
--- a/packages/button/src/Button/Button.tsx
+++ b/packages/button/src/Button/Button.tsx
@@ -13,7 +13,7 @@ import { useStyleEngine } from "@telegraph/style-engine";
 import { Text as TelegraphText } from "@telegraph/typography";
 import clsx from "clsx";
 import { LoaderCircle } from "lucide-react";
-import React, { ButtonHTMLAttributes } from "react";
+import React from "react";
 
 import {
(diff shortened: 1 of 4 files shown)


packages/button/src/Button/Button.tsx
```diff
@@ -35,7 +35,6 @@ type RootBaseProps = {
   size?: ButtonSize;
   state?: "default" | "loading";
   active?: boolean;
-  type?: "button" | "submit";
```
andy-knock, line 38: discussed with @kylemcd live this morning. Do not need to type this prop here when we can already just unpack it on line 85

PR #773, opened 2026-04-16
fix(KNO-12734): add portal around menu content

kylemcd:
## What changed

- Wrapped `Menu.Content` in `RadixMenu.Portal` so the menu renders outside the DOM subtree.
- Added `className="tgph"` to the content `Stack` so styles apply correctly in portals.
- Removed the now-redundant `Portal.Root asChild` wrapper from `Combobox.Content` — `Menu.Content` already portals itself, so Combobox was double-portaling to `document.body`.
- Added type-level tests for `MenuContentProps` to match the pattern in `Popover`.

## Why

Menu content was not rendering in a portal, which caused Safari rendering bugs. This also made it inconsistent with the `Popover` component, which already uses `RadixPopover.Portal`. Both components now behave the same way.

The `Portal.Root asChild` in Combobox was originally added as a workaround for the missing portal in `Menu.Content`. With the portal now living inside `Menu.Content`, that wrapper became a no-op double portal and has been removed.

## Notes

The `tgph` class is required whenever content renders in a portal — without it, Telegraph's scoped CSS does not apply to the content. This mirrors the approach already used in `Popover.tsx`.

### DIFF
--- a/.changeset/fast-zebras-fetch.md
+++ b/.changeset/fast-zebras-fetch.md
@@ -1,5 +1,6 @@
 ---
 "@telegraph/menu": patch
+"@telegraph/combobox": patch
 ---
 
-fix(KNO-12734): Wrap menu content in a portal to fix Safari rendering issues and match Popover component behavior
+fix(KNO-12734): Wrap menu content in a portal to fix Safari rendering issues and match Popover component behavior; remove resulting double portal from Combobox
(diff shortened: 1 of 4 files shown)


kylemcd:
* **#773** <a href="https://app.graphite.com/github/pr/knocklabs/telegraph/773?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/knocklabs/telegraph/773?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

connorlindsey (approved):

PR #759, opened 2026-04-13
chore(deps): upgrade vite to v8 and @vitejs/plugin-react to v6

kylemcd:
## What changed

- Upgraded `vite` from `^6.4.1` to `^8.0.0` across all 28 component packages and `@telegraph/vite-config`.
- Upgraded `@vitejs/plugin-react` from `^5.1.2` to `^6.0.0` in `@telegraph/vite-config`.
- Renamed `build.rollupOptions` to `build.rolldownOptions` in `vite-config.ts` and `style-engine-config.ts` — `rollupOptions` is deprecated in Vite 8 in favor of the rolldown-backed equivalent.
- Updated the `PreRenderedAsset` type import in `vite-config.ts` from `rollup` to `vite`'s re-exported `Rollup` namespace, since Vite 8 uses rolldown internally and no longer exposes rollup types directly.
- Added `build.cssTarget` to `vite-config.ts` to configure LightningCSS (Vite 8's CSS minifier) with appropriate browser targets, preserving `-webkit-appearance` and `-webkit-user-select` vendor prefixes that autoprefixer emits for iOS < 15.4.
- Bumped `@vitest/coverage-v8` from `4.1.0` to `4.1.4` to match the already-current `vitest` version.
- Removed the stale `vite >= 7.0.0` ignore rule from `.github/dependabot.yml`.

## Why

The repo had been blocked from upgrading past Vite 6 by a Dependabot ignore rule added when Storybook didn't support Vite 7+. That blocker is gone — `@storybook/react-vite` added Vite 7 support in v10.2.0 and Vite 8 support in v10.3.0, and the repo is already on `^10.3.1`. The Dependabot rule was stale and preventing us from getting security and performance improvements in Vite 8.

`@vitejs/plugin-react` v6 is coupled with the Vite 8 upgrade since it declares `vite: ^8.0.0` as a peer dependency.

Vite 8 switched from esbuild to LightningCSS for CSS minification. LightningCSS derives its vendor prefix behavior from `build.cssTarget`. Without an explicit target, it defaults to a modern baseline and omits prefixes that autoprefixer was adding. Setting `cssTarget` to include iOS 13 and Safari 13 restores the `-webkit-appearance` and `-webkit-user-select` prefixes. The `-moz-appearance` and `-moz-user-select` prefixes are correctly dropped — Firefox 80+ supports the unprefixed forms.

## Notes

- All 31 package builds pass and all 295 tests pass against Vite 8.
[...]

PR #827, opened 2026-06-08
fix(vite-config): externalize Node built-in modules so style-engine's PostCSS plugin works

kylemcd:
## Summary

`@telegraph/style-engine`'s PostCSS plugin loads Node built-ins via `require("node:path")` / `require("node:fs")`. The shared Vite build config (`packages/vite-config`) only externalized each package's declared dependencies — never Node built-ins.

Under Vite 7 / Rollup these were auto-externalized, so `style-engine@0.3.2` shipped correctly. [#759](https://github.com/knocklabs/telegraph/pull/759) upgraded to Vite 8 / Rolldown, which does **not** auto-externalize `node:` built-ins here — it inlined the requires as empty-object stubs. At runtime `path`/`fs` were `{}`, so `path.dirname` was `undefined` and the plugin threw:

```
TypeError: t.dirname is not a function
  at .../node_modules/@telegraph/style-engine/dist/cjs/postcss.js
```

`0.3.3` and `0.3.4` were the first releases rebuilt under the new config, so the break first appeared there. This is currently blocking [control#9558](https://github.com/knocklabs/control/pull/9558) (telegraph dependabot group bump).

## Fix

Externalize Node built-ins (bare and `node:`-prefixed) in `packages/vite-config/src/vite-config.ts` by adding `builtinModules` to the Rolldown `external` list. This is the correct general fix; in practice `style-engine` is the only package that ships built-in `require`s in its runtime output, so it is the one that needs republishing.

The changeset bumps both `@telegraph/vite-config` (the fix) and `@telegraph/style-engine` → `0.3.5` (forcing a republish under the fixed config, since `vite-config` is only a devDependency and would not otherwise cascade).

Once `0.3.5` publishes, control#9558 picks it up and Vercel goes green — no control-side change needed.

## Verification

After rebuilding under the fixed config, `packages/style-engine/dist/cjs/postcss.js` emits real `require("node:path")` / `require("node:fs")` with no empty-object stub:

```js
var e=require(`node:path`),t=require(`node:fs`),n=require(`postcss`);
```

Resolves [KNO-13579](https://linear.app/knock/issue/KNO-13579).

### DIFF
--- a/.changeset/kno-13579-vite-config-externalize-node-builtins.md
+++ b/.changeset/kno-13579-vite-config-externalize-node-builtins.md
@@ -1,0 +1,16 @@
+---
+"@telegraph/vite-config": patch
+"@telegraph/style-engine": patch
+---
+
+fix(vite-config): externalize Node built-in modules in the Rolldown bundle
+
+The shared Vite build config only externalized each package's declared
+dependencies, never Node built-ins. Under Vite 8 / Rolldown, runtime
+`require("node:path")` / `require("node:fs")` calls were inlined as empty-object
+stubs, so `@telegraph/style-engine`'s PostCSS plugin threw
+`TypeError: t.dirname is not a function` and broke every consumer build.
+
+Add `builtinModules` (bare and `node:`-prefixed) to the Rolldown `external`
+list. Republishes `@telegraph/style-engine` (0.3.5) so its `dist/cjs/postcss.js`
+emits real `require("node:path")` / `require("node:fs")` again.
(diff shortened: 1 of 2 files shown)


kylemcd:
* **#827** <a href="https://app.graphite.com/github/pr/knocklabs/telegraph/827?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/knocklabs/telegraph/827?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

meryldakin (approved):

PR #864, opened 2026-07-08
fix(base-ui): stack Tabs vertically and composite Modal enter animation

kylemcd:
Fixes two visual regressions from the Base UI migration ([#837](https://github.com/knocklabs/telegraph/pull/837)), both spotted on infra-dev while verifying control's dep bump.

## What changed

**`@telegraph/tabs` — tab list rendered beside the panel instead of above it.**
The root now renders a vertical `Stack` (`direction="column"`) instead of a bare `Box`. Consumers previously had to pass `as={Stack} direction="column"` to get the standard layout — but that combination is broken: `Box` maps `direction` to `--flex-direction` while the `Stack` CSS class reads `--direction`, so the var was ignored and the list fell back to a full-height `row`. The root now owns its layout; `direction` is still overridable for side-by-side tab layouts.

**`@telegraph/modal` — enter animation snapped into place on busy pages.**
The slide-in now animates a GPU-composited `transform: translateY(...)` instead of the `top` layout property, with the resting position set statically. `top` is not composited, so every animation frame forced layout + paint on the main thread; on a busy page (e.g. the create-workflow modal opening over a ~285-row list) the spring dropped frames and snapped. An idle Storybook page never stressed it, which is why it only reproduced in control. Also settled the stack-depth scale at exactly `1` for a lone modal (it rested at `1.02`, ~2% too large).

## Why it only showed in control

The scale math regressed because the migration changed the unresolved-layer fallback from `-1` to `0`, and the snap came from `top` being non-composited — both only surface under a real, busy render. Extracted the scale into a pure `getStackedModalScale` helper with unit tests, and added Tabs tests asserting the vertical layout.

## Verification

- Build, lint, and all tabs + modal unit tests green.
- Verified live in Storybook: Tabs root computes `flex-direction: column`; Modal enter animates `transform: translateY(-16px → 0)` with `top` static at `64px`, and the content settles at `scale(1)` (no residual `1.02`).

Resolves [KNO-14078](https://linear.app/knock/issue/KNO-14078)

### DIFF
--- a/.changeset/fix-tabs-vertical-modal-scale.md
+++ b/.changeset/fix-tabs-vertical-modal-scale.md
@@ -1,0 +1,9 @@
+---
+"@telegraph/tabs": patch
+"@telegraph/modal": patch
+---
+
+Fix two visual regressions from the Base UI migration:
+
+- **Tabs**: the root now renders a vertical `Stack` (`direction="column"`) by default so the tab list sits above the tab panel. Previously it rendered a bare `Box`, forcing consumers to pass `as={Stack} direction="column"` — which silently fell back to a horizontal `row` layout because `Box` emits `--flex-direction` while the `Stack` CSS reads `--direction`. Consumers can still override `direction` for side-by-side layouts.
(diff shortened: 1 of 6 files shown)


kylemcd:
* **#864** <a href="https://app.graphite.com/github/pr/knocklabs/telegraph/864?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/knocklabs/telegraph/864?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

cjbell (approved):

PR #868, opened 2026-07-08
fix(layout): share --direction custom property between Box and Stack

kylemcd:
## What changed

- **`@telegraph/layout`**: `Box` now maps `direction`/`flexDirection` to `--direction` — the same custom property `Stack` uses — and `.tgph-box` defaults and reads it; interactive pseudo-state fallbacks regenerate as `--hover--direction` etc.
- **`layout` tests**: regression coverage for `<Box as={Stack} direction="column">`, the `flexDirection` alias on both components, and a parity test asserting Box and Stack agree on the custom property.
- **`tabs` tests**: `<Tabs.Panel as={Stack} direction="column">` keeps its column layout.
- Patch changeset for `@telegraph/layout`.

## Why

`Box` wrote `--flex-direction` while the `.tgph-stack` rule reads `--direction`. On an element with both classes — any Box-based component composed with `as={Stack}` — the stack rule wins the cascade and resolves `flex-direction` from the never-set `--direction`, so `direction="column"` silently rendered a row. This is the root cause behind the tab layouts breaking in control (worked around with inline styles in knocklabs/control#10137); #864 fixed the Tabs root by making it a real `Stack`, but `TabPanel` and every other Box-based component were still affected.

## Notes

Ships as a patch — the old `--flex-direction` var was already dead on any element with the stack class, and nothing references it anymore.

Resolves [KNO-14080](https://linear.app/knock/issue/KNO-14080/telegraph-layout-boxstack-flex-direction-var-mismatch-drops-direction)

### DIFF
--- a/.changeset/fix-box-stack-direction-var.md
+++ b/.changeset/fix-box-stack-direction-var.md
@@ -1,0 +1,7 @@
+---
+"@telegraph/layout": patch
+---
+
+Fix `Box` and `Stack` using different CSS custom properties for `flex-direction`, which silently dropped the `direction`/`flexDirection` prop when composing Box-based components with `as={Stack}`.
+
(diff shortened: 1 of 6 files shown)


kylemcd:
* **#868** <a href="https://app.graphite.com/github/pr/knocklabs/telegraph/868?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/knocklabs/telegraph/868?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

cjbell (approved):

PR #869, opened 2026-07-08
fix(menu): keep focus on a typeable trigger input when the popup opens

kylemcd:
## What changed

`MenuContent`'s prevented-`onOpenAutoFocus` path no longer restores focus in a `setTimeout(0)`. It now bounces focus back to the intended target synchronously from a one-shot `focusin` listener on the popup — the instant Base UI's initial focus lands. The bounce disarms after the first restore, on the first `pointerdown`/`keydown` inside the popup, and on close, so navigation and clicks into the menu still work.

## Why

Since the Base UI migration, an input composed inside `Menu.Trigger` couldn't be typed into: the popup stole focus on open, so keystrokes fed the menu typeahead instead of the input (breaks PropertySelectorField everywhere it appears, plus the block editor slash-command / variable / skill-suggestion menus). Base UI's `MenuPopup` hardcodes its initial focus and queues it on the next animation frame, so the old `setTimeout(0)` restore fired first and then lost the race — the popup won every time. Bouncing from `focusin` runs inside Base UI's own `.focus()` call, so it wins deterministically with no timer assumptions.

## Tests

- **jsdom unit tests** assert the bounce *logic* via a spy on the restore target's `focus()` — they fail against the old `setTimeout` code. jsdom can't assert the real outcome because it re-asserts `activeElement` after a `.focus()` returns, clobbering the synchronous re-focus.
- **Real-browser tests** (`Menu.browser.test.tsx`, Vitest Browser Mode + Playwright, opt-in via `yarn workspace @telegraph/menu test:browser`) drive the actual typeable-trigger flow — including the exact `TypeableTrigger` Storybook composition (shared via `Menu.fixtures.tsx`) — and assert focus stays on the input while typing filters the list. They run **headless by default** (the fix is deterministic, so no window pops up); `test:browser:headed` reproduces the race itself.
- There's a **`TypeableTrigger` story** back in Storybook to manually test the input, verified by the browser test above so it can't silently break.
- **The race only reproduces in a headed browser.** Base UI defers its initial popup focus to a real animation frame; every headless env — jsdom, happy-dom, *and headless Chromium* — fires it eagerly, so the buggy code passes headless. Run headed to watch the old `setTimeout` restore lose (focus jumps to the popup). A headed CI job (xvfb) and generalizing the harness are tracked in [KNO-14089](https://linear.app/knock/issue/KNO-14089).

## Notes

- Upstream fix (not blocking): [mui/base-ui#5189](https://github.com/mui/base-ui/pull/5189) adds an `initialFocus` prop to `Menu.Popup`. If the Base UI team accepts it, we can delete the bounce and map a prevented `openAutoFocus` to `initialFocus === false` — the same mapping Telegraph's `Popover` and `Modal` already use — bringing `Menu` in line with the other primitives.

Resolves [KNO-14086](https://linear.app/knock/issue/KNO-14086/telegraph-menu-openautofocus-restore-races-base-ui-initial-focus)

### DIFF
--- a/.changeset/menu-open-autofocus-focus-bounce.md
+++ b/.changeset/menu-open-autofocus-focus-bounce.md
@@ -1,0 +1,7 @@
+---
+"@telegraph/menu": patch
+---
+
+Fix a focus race from the Base UI migration where an input composed inside `Menu.Trigger` could not be typed into: the menu popup stole focus on open, so keystrokes fed the menu typeahead instead of the input (breaking PropertySelectorField everywhere it appears, plus the block editor slash-command and variable suggestion menus).
+
+`MenuContent`'s `onOpenAutoFocus` shim emulated Radix's prevented-autofocus contract by re-asserting the intended focus in a `setTimeout(0)`. Base UI's `MenuPopup` hardcodes its initial focus and queues it on the next animation frame, so the `setTimeout(0)` restore fired first and then lost the race — the popup won every time. The shim now bounces focus back synchronously from a one-shot `focusin` listener on the popup: when Base UI's queued initial focus lands, the listener re-focuses the intended target during that same `.focus()` call, with no dependence on timer ordering. The bounce disarms after the first restore, on the first `pointerdown`/`keydown` inside the popup, and on close, so ArrowDown navigation and item clicks still move focus into the menu.
--- a/.gitignore
+++ b/.gitignore
@@ -51,3 +51,7 @@ debug-storybook.log
 # moon
 .moon/cache
 .moon/docker
+
+# Vitest browser mode artifacts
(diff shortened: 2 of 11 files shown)


kylemcd:
* **#869** <a href="https://app.graphite.com/github/pr/knocklabs/telegraph/869?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/knocklabs/telegraph/869?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

MikeCarbone (approved):
tested and does this thing it says it does

PR #873, opened 2026-07-09
fix: override rounded corners in combobox option tag

BenaventeX24:
### WHAT CHANGED
pass down correct border values to Tag component from Combobox so that we fix: (screenshots)

### BEFORE:
image

### NOW:
image

### DIFF
--- a/packages/combobox/src/Combobox/Combobox.primitives.tsx
+++ b/packages/combobox/src/Combobox/Combobox.primitives.tsx
@@ -350,6 +350,8 @@ const TriggerTagButton = <T extends TgphElement>({
     <Tag.Button
       icon={{ icon: X, alt: `Remove ${triggerTagContext.value}` }}
       height="full"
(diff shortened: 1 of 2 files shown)


packages/combobox/src/Combobox/Combobox.primitives.tsx
```diff
@@ -350,6 +350,8 @@ const TriggerTagButton = <T extends TgphElement>({
     <Tag.Button
       icon={{ icon: X, alt: `Remove ${triggerTagContext.value}` }}
       height="full"
+      borderTopRightRadius="1"
+      borderBottomRightRadius="1"
```
kylemcd: this can just be `borderRightRadius="1"` instead

BenaventeX24: ABSOLUTELY RIGHT many thanks

kylemcd (approved):

PR #883, opened 2026-07-16
fix(combobox): correct client-side search matching (KNO-14268)

kylemcd:
## What changed

`Combobox`'s client-side search read option text off the React element tree, so it missed text it renders on screen — most visibly, text produced inside a child component.

**Before** — typing `kyle@knock.app` filters out the option showing that email:

before

**After** — the same option matches, with no consumer changes:

after

- Matches options by their rendered DOM text, captured per option while it's on screen. Text inside wrapper components (like a custom user row) is now searchable automatically.
- Normalizes whitespace on the query and option text.
- Matches across sibling nodes, and collects numeric children.
- Stops treating a controlled `<Combobox.Search value>` (or any controlled input inside `Content`) as an option.

## Why

Reported as email search breaking after the Base UI migration. Not a regression, these are long-standing bugs. Only affects an uncontrolled `<Combobox.Search />` — server-side search bypasses this filter.

Resolves KNO-14268

### DIFF
--- a/.changeset/combobox-search-matching.md
+++ b/.changeset/combobox-search-matching.md
@@ -5,7 +5,7 @@
 Fix client-side search matching in `Combobox`
 
 - Match options by their rendered DOM text, so text produced inside a child component (a custom option row, for example) is searchable. Previously the filter only read text off the React element tree, where a child component's render output doesn't exist.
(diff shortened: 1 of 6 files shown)


MikeCarbone (approved):
nice!

PR #185, opened 2024-07-15
fix: combobox height issues

kylemcd:
### Description 
- Fixes height issues seen with the combobox trigger
- Adds min height and min width as prop values for layout components
- Fixes incorrect reference to flex wrap in Stack and adds shorthand

### Tasks
[KNO-6406](https://linear.app/knock/issue/KNO-6406/[telegraph]-fix-height-issue-with-combobox)

PR #887, opened 2026-07-17
fix(truncate): stop useTruncate effect re-running on every render

kylemcd:
Resolves [KNO-14285](https://linear.app/knock/issue/KNO-14285/maximum-update-depth-crash-in-workflow-graph-editor-from)

## What changed

- **`TooltipIfTruncated`** now keys `useTruncate`'s measurement effect on the child's stable rendered **text** (`getTextContent(children)`) instead of the freshly-created `children` element.
- **`useTruncate`** only writes `truncated` state when the value actually changes (`setTruncated(prev => (prev === next ? prev : next))`).

## Why

The dashboard was throwing React error #185 ("Maximum update depth exceeded") from `useTruncate`, rendered inside the ReactFlow-based Workflow Graph Editor step cards (`BaseStepCard` title).

`children` is a brand-new element on every render, so keying the effect on `[children]` re-mounted it — and called `setState` — on **every** render. That's harmless while React's same-value `setState` bailout absorbs it, but in a continuously re-rendering tree (ReactFlow) the fiber's lanes stay dirty, defeating the bailout, and the passive `setState` cascades into a runaway synchronous update loop past React's 50-nested-update limit.

Two independent fixes, either of which breaks the loop; together they're belt-and-suspenders:
1. Stable text dependency → the effect stops re-mounting on every render.
2. Value-guarded `setState` → a redundant write is a genuine no-op even when lanes are dirty.

The Base UI tooltip migration (#10170) is what tipped this latent defect into a crash, but the root cause lives in `@telegraph/truncate`, so it's fixed here.

## Testing

- **New regression test** (`TooltipIfTruncated.test.tsx`): 10 parent re-renders with fresh-but-identical children now construct exactly **1** `ResizeObserver` — before the fix it was 11 (one per render, matching the ticket's evidence). Verified the test fails against the old `[children]` dep.
- `@telegraph/truncate` jsdom (11) + browser (2) tests, lint, prettier, and typecheck all pass. Real-overflow tooltip behavior is unchanged.
- Changeset included (`@telegraph/truncate` patch).

## Follow-up

Bump `@telegraph/truncate` in the dashboard once this patch releases.

### DIFF
--- a/.changeset/truncate-max-update-depth-fix.md
+++ b/.changeset/truncate-max-update-depth-fix.md
@@ -1,0 +1,7 @@
+---
+"@telegraph/truncate": patch
+---
+
+fix(truncate): prevent "Maximum update depth exceeded" crash in re-render-heavy environments
+
+`TooltipIfTruncated` keyed `useTruncate`'s measurement effect on the freshly-created `children` element, so the effect re-mounted and called `setState` on every render. In continuously re-rendering trees (e.g. a ReactFlow graph editor) React's same-value bailout could be defeated, cascading into a runaway synchronous update loop (React error #185). The effect now depends on the stable rendered text instead of the element identity, and `useTruncate` only writes state when the truncation value actually changes.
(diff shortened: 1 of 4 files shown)


kylemcd:
* **#887** <a href="https://app.graphite.com/github/pr/knocklabs/telegraph/887?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/knocklabs/telegraph/887?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #895, opened 2026-07-21
fix: tighten base-ui dismiss handler types so stale Radix usage fails to compile

kylemcd:
## What changed

- **`Popover.Content` / `Menu.Content`** — dismiss-handler `event` params (`onInteractOutside`, `onPointerDownOutside`, `onFocusOutside`, `onEscapeKeyDown`) now resolve to their concrete DOM `Event` at the JSX call site instead of `any`; polymorphic props come from `PolymorphicProps<T>` + non-generic `Omit<StackProps, "align" | "as">` so the `align`-Omit no longer wraps the generic `Stack<T>`.
- **`Modal.Content`** — same fix, converted from `forwardRef` to a `tgphRef` function component (its `PropsWithoutRef` wrapper caused the same widening).
- **`*.test.tsx`** — compile-time guards that inline handler params are concrete `Event`s and that stale `event.detail` access fails to compile.

## Why

Telegraph's base-ui shim keeps Radix's dismiss-handler prop names but hands the callback the native DOM event. The param was silently widening to `any` at the JSX call site, so a stale Radix `event.detail.originalEvent` handler compiled and then crashed at runtime (`Cannot read properties of undefined`, Sentry `CONTROL-DASHBOARD-2H7`). jsdom can't exercise real outside-press, so the type-level guard is the durable prevention.

### DIFF
--- a/.changeset/tidy-dismiss-handler-types.md
+++ b/.changeset/tidy-dismiss-handler-types.md
@@ -1,0 +1,11 @@
+---
+"@telegraph/popover": patch
+"@telegraph/menu": patch
+"@telegraph/modal": minor
+---
+
+Tighten base-ui dismiss-handler types so stale Radix `event.detail` usage fails to compile.
+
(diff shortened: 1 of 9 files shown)


MikeCarbone (approved):
tested storybook preview and looks good

PR #919, opened 2026-07-27
fix(toggle): keep the track color, dimmed, when a Toggle is disabled

kylemcd:
Resolves [KNO-14456](https://linear.app/knock/issue/KNO-14456/fixkno-upstream-a-disabled-toggle-renders-the-same-whether-it-is-on-or)

## What changed

- A disabled track keeps its own color at 50% opacity: the `color` token when on, `--tgph-gray-7` when off.
- Dropped the `:not([data-tgph-button-state="disabled"])` guard from the unchecked-track rule, so the off state is dimmed gray rather than Button's flat gray.
- Added `DisabledOn` / `DisabledOff` stories.
- Added `Toggle.browser.test.tsx` covering the four checked × disabled combinations.

## Why

`Toggle.Switch` renders its track as a `Button.Root`, and Button paints every disabled solid track `--tgph-gray-3`. Toggle's own track rule explicitly excluded the disabled case, so disabled+on and disabled+off rendered the identical gray — thumb position and a gray icon were the only cue. Chakra composed `_disabled: { opacity: 0.4 }` with the checked background, so a locked switch showed a faded blue track.

The fix leans on `Button.Root` still setting the inline `--background-color` var to the color token when disabled; only its blanket rule was outranking it. Reading that var back is what keeps this a toggle-scoped patch instead of duplicating `BUTTON_COLOR_MAP`. `packages/menu/src/default.css:16` already reaches for the same var across a package boundary.

## Notes

- **Off state changes appearance too**, from flat `--tgph-gray-3` to `--tgph-gray-7` at 50%. Over white that lands close to where it was, so existing off+disabled toggles should look about the same.
- **Why a browser test.** The bug is entirely cascade arbitration. jsdom doesn't resolve `var()`, so no jsdom assertion can go red on it — verified by reverting the fix and watching 2 of 3 tests fail. `Toggle.browser.test.css` exists to control stylesheet load order: Toggle's sheet loads first so it loses every source-order tie with Button's equally specific rule, which makes the `!important` load-bearing in the test the way it is in production. Removing just the `!important` also reddens it.
- Deps' CSS is imported from `dist`, so the browser job needs `yarn build:packages` first. CI already does this.
- **Two changes beyond the fix**, both easy to drop if you'd rather:
  - `vitest.config.mts` now merges `sharedConfig` — required, it's what keeps `*.browser.test.tsx` out of the jsdom run. Matches `menu`/`modal`/`truncate`.
  - `tsconfig.json` now excludes test files — optional convention alignment (20 of 31 packages, including every package with a browser test). Without it the package reports type errors on its own test files, ~20 of which predate this PR.
- The `Toggle.Indicator` tag is not dimmed alongside the track, so a disabled toggle shows a full-strength "Enabled" tag next to a faded track. Left alone as a design call rather than folded in here.

## Testing

`yarn test` 551 passing, `yarn test:browser` 13 passing, lint/format/`manypkg` clean. Verified light and dark in Storybook.

### DIFF
--- a/.changeset/wild-hounds-taste.md
+++ b/.changeset/wild-hounds-taste.md
@@ -1,0 +1,12 @@
+---
+"@telegraph/toggle": patch
+---
+
+fix(toggle): keep the track color, dimmed, when a Toggle is disabled
+
+A disabled Toggle inherited Button's blanket disabled background, so on and off
+rendered the identical `--tgph-gray-3` and thumb position was the only cue. A
+disabled track now keeps its own color — the `color` token when on, `--tgph-gray-7`
+when off — at 50% opacity, so a locked toggle still reads as on or off. This
+changes the appearance of the disabled off state as well, from a flat
+`--tgph-gray-3` to the dimmed unchecked gray.
--- a/packages/toggle/package.json
+++ b/packages/toggle/package.json
@@ -45,6 +45,7 @@
     "@knocklabs/typescript-config": "^0.0.2",
     "@telegraph/postcss-config": "workspace:^",
     "@telegraph/prettier-config": "workspace:^",
+    "@telegraph/tokens": "workspace:^",
     "@telegraph/vite-config": "workspace:^",
     "@types/react": "^19.2.17",
     "eslint": "^10.4.0",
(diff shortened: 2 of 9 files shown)


kylemcd:
* **#919** <a href="https://app.graphite.com/github/pr/knocklabs/telegraph/919?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/knocklabs/telegraph/919?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #962, opened 2026-08-12
chore(repo): pin local Node to 24.19.0 to match CI

kylemcd:
## What changed

`.tool-versions` now says `nodejs 24.19.0`.

## Why

`.tool-versions` asked for Node 22 while `engines` asks for `24.x`, and every workflow reads `engines` via `node-version-file: "package.json"`. Local development ran a different major than CI.

The 22 pin also moved to 22.23.2 in #950 to satisfy jsdom 30 (`^22.22.2 || ^24.15.0 || >=26.0.0`). Anyone without that exact version installed via asdf gets `No version is set for command node` and cannot run `yarn` at all. 24.19.0 satisfies jsdom through its `^24.15.0` range instead, and matches what CI already resolves.

## Notes

Verified on 24.19.0 from a clean install: `yarn install`, `build:packages`, 927 tests with no type errors, eslint and prettier all pass.

Tooling-only, so the changeset is empty and nothing publishes.


kylemcd:
Folded into #955 — the Node pin belongs with the type-check work rather than on its own.

